PR #5, opened 2026-01-08
Deduplicate window resources

t4st20-hue:
Vibecoding can get things done fast, but it sure does result in a lot of deeply unmaintainable stuff.

There are duplicate bits of code EVERYWHERE, all sorts of buttons, and way too many hard coded colors. Moving them to a central location elsewhere will make it easier to adjust the behavior or look later, so I have started that effort!

TODO: There are a bunch more random hardcoded colors I need to replace with brushes.....
TODO: The actual button code in controls.xaml can probably be shared much more between types.....


CodeBambi:
Really nice cleanup work here. Centralizing the theme resources makes so much more sense than having the same PinkBrush defined in every single window.

The structure looks clean - Colors -> Brushes -> Controls is a logical hierarchy. And getting rid of ~500 lines of duplicate code is always a win.

I noticed you mentioned there are still more hardcoded colors to tackle. No rush on that - this is already a solid improvement as-is. Can always do another pass later.

Good to merge! ❤️

PR #25, opened 2026-05-28
feat: Add OpenAI-compatible AI provider support

ObviouslyNotMich:
- Introduced a new AI provider type for OpenAI-compatible services.
- Updated UI to allow selection of the OpenAI-compatible provider.
- Implemented configuration fields for OpenAI endpoint, API key, and model.
- Added daily request limit functionality for OpenAI-compatible provider.
- Created OpenAiCompatibleService to handle requests to OpenAI-compatible endpoints.
- Integrated secure string handling for API key storage.
- Updated localization files to include new labels for OpenAI-compatible settings.
- Refactored existing AI service strategy to accommodate new provider type.

image

Idea is to be able to use remotely hosted public uncensored models they go way harder than the normal guide railed ones. 
See: https://nano-gpt.com/models/text

Sort by uncensored


ObviouslyNotMich:
3:00 am here so any bugs or style issues ill fix tomorrow or Friday feel free to comment :) 

Edit: but from testing it all for 30 min it seems to work.

ObviouslyNotMich:
@CodeBambi Made a bunch of changes. Also fixed the pink filter and added a alternative rendering option that should be faster with d3d. It will go back to legacy if it see on launch that it failed once.

CodeBambi:
hey, finally got around to properly reviewing this, sorry for the wait.. last weeks have been hectic on my side.

The provider itself looks solid, diagnostics + the daily limit are a nice touch. Two things before i can merge:

1. Could you split the D3D overlay stuff into its own PR? It looks intresting but it's a big change to the rendering path and I want to test it seperately. With just the provider files this is basically mergeable (only .gitignore and FlashService conflict with current main)
2. Right now replies from this provider skip the effects parser, so OpenAI-compatible models can't trigger panel effects like local/cloud can. Fine for v1 if you note it, but worth a follow up.

Rebase on main when you get a chance and ill take another look :)

PR #33, opened 2026-06-05
feat(bark): PR1 — bark system loader + dry-run matcher (no speak path)

CodeBambi:
## What this is

PR1 of the reactive companion-dialogue (\"bark\") system. **Infrastructure only — nothing is spoken yet.** It loads rule data, wires the full event-subscription seam, and logs which bark *would* fire and why. This is the safe-to-merge foundation before the speak path lands.

## What's included

**New — `Services/Bark/` + `Services/BarkService.cs`:**
- `BarkRule` / `BarkScope` / `BarkClass` — JSON rule data model (`{ trigger, conditions, priority, cooldown_ms, repeatable, scope, mood, class, variant_pool | pool_ref }`).
- `BarkRuleSet` + `BarkRuleLoader` — two-tier load (embedded base manifest + active-mod overlay, merged by `id`), mirroring how `CompanionPhraseService` resolves mod-vs-embedded resources.
- `BarkContext` — per-fire value bag handed to the matcher.
- `BarkState` — in-memory counter layer for conditions the app doesn't expose as events: cumulative blink tally, prolonged face-lost timer, rapid mod-switch window, days-away / instant-relaunch (from `AppSettings.LastSeenUtc`), marathon thresholds.
- `BarkService` — `GamificationBridge`-style `Start()/Stop()` with one subscription block over the confirmed event surface; `AttachSessionEngine`/`AttachTray` for the MainWindow-owned, lazily-created `SessionEngine` and `TrayIconService`; a **dry-run matcher** that filters by conditions and reuses the existing cooldown / global-min-gap / one-fire-latch / variant-rotation primitives, logging every decision.

**Wiring:**
- `App.xaml.cs` — construct `Bark` beside `Gamification`; `Start()` after the bridge.
- `MainWindow.xaml.cs` — `AttachTray` at tray init; `AttachSessionEngine` when the engine is created.

**Sample manifest** (`Resources/sounds/companion_audio/bark_rules.json`) — placeholder rules so dry-run is demonstrable. Real line content arrives separately as data.

## How to verify

Run the app and watch the log. Trigger events (start a session, level up, fail a video attention check, etc.) and look for:
```
[BARK dry-run] WOULD FIRE trigger=SessionStarted rule=sample_session_greeting class=Normal mood=warm priority=10 variant#=0 line=\"There you are. Let's begin.\" gate=OK
[...]

PR #34, opened 2026-06-05
feat(bark): PR2 — speak path (output, safety, chat-suppression, self-echo, Fork D)

CodeBambi:
## What this is

PR2 of the reactive bark system: the **speak path**. Turns PR1's dry-run matcher into a real speaker while keeping the dry-run toggle for validation.

> **Stacked on #33 (PR1).** Base is `feat/bark-service-pr1`, so this diff is just the speak-path delta. Merge #33 first (or retarget this to `main` after #33 lands).

> ⚠️ **Merge is pending your PR1 dry-run validation pass.** PR1 #33 now also carries two verification fixes (field-level manifest merge + idempotent `AttachTray`) — please run the dry-run validation against #33 before merging this.

## Included

- **`Fire()` speaks** — routed by class/priority: non-Normal class *or* `priority >= 100` preempt via `AvatarTubeWindow.GigglePriority(line, aiGenerated:false)`; ordinary barks queue via `Giggle`. The decision is taken under the lock and the render happens after releasing it (Giggle/GigglePriority marshal to the UI thread). `{0}` (focused app) and `{key}` (context) substitution applied at speak time.
- **Safety short-circuit** — `class=safety` bypasses the gate; `GigglePriority` clears the queue; a fired safety bark holds the floor for `SafetyHoldMs` so no non-safety bark fires over it. (No trigger raises a safety rule yet — the machinery is correct for when a panic/abort event maps to a safety-class rule.)
- **Chat-suppression (Fork E)** — new `AppSettings.BarkChatSuppressionMs` (default **10000**). Non-safety barks are suppressed while the companion is busy (`_isWaitingForAi` / `_isShowingAiBubble`, via new `AvatarTubeWindow.IsCompanionBusy(windowMs)`) or within the window after the last `UserMessageSent` / AI reply.
- **Self-echo mute** — new public `KeywordTriggerService.MuteKeywordEcho(text, muteMs)` (thin wrapper over the private `ForceMuteKeyword`), called after each bark renders so the bubble text can't trip awareness/OCR.
- **Persistence (Fork B)** — `AppSettings` gains `NewYearNoteReactionSeen`, `BarkLifetimeFired` (+ `MarkBarkFired`/`IsBarkFired`). once/lifetime and once/tier latches persist (tier key is `id@Tier`, so a tier change re-arms); session-scope latches stay in-memory. **Dry-run never writes persisted latches.**
- **Lock card 50/50 (Fork D)** — `BarkService` is the **sole** `LockCardCompleted` subscriber and does one coin flip per event: heads → `await AvatarTubeWindow.PlayLockCardAiReactionAsync(e)` (the old self-firing handler, now public, returns whether it spoke); tails → guaranteed pool bark. Heads with AI unavailable *or* AI silent falls through to the pool bark, so **exactly one** reaction always fires, never both/neither. `AvatarTubeWindow` no longer self-subscribes (subscription + cleanup unsubscribe removed).

## DryRun toggle
`DryRun` defaults to **false** (PR2 speaks). Run with **`CCP_BARK_DRYRUN=1`** for log-only validation — the full matcher runs and logs `[BARK dry-run] WOULD FIRE … / blocked … reason=…` without rendering to the avatar or writing persisted latches.

⚠️ Note: with real speech on, the **sample placeholder lines** in `bark_rules.json` will actually be spoken. They're stand-ins until you supply real line content.

## Explicitly deferred to the egg PR (as planned)
- Mute-egg `MasterVolume==0` guard before `PlayGiggleSound`.
- `PlayNoteClip` real-recording playback + the logo-egg (100-clicks-in-60s) wiring at `MainWindow.ShowEasterEgg()`.
- The lifetime-persistence fields land **now** because the speak path needs them; the egg wiring consumes `NewYearNoteReactionSeen` later.

## Build
0 errors, no new warnings in the touched files.

### DIFF
--- a/ConditioningControlPanel/AvatarTubeWindow.xaml.cs
+++ b/ConditioningControlPanel/AvatarTubeWindow.xaml.cs
@@ -379,8 +379,9 @@ namespace ConditioningControlPanel
                 App.Video.VideoEnded += OnVideoEnded;
             }
 
-            if (App.LockCard != null)
-                App.LockCard.LockCardCompleted += OnLockCardCompleted;
+            // Lock-card reaction is now owned by BarkService (Fork D 50/50 coin flip): it is the
+            // sole subscriber to LockCardCompleted and invokes PlayLockCardAiReactionAsync on heads.
+            // This window no longer self-subscribes.
 
             // Wire up game completion events
             if (App.BubbleCount != null)
@@ -2022,8 +2023,7 @@ namespace ConditioningControlPanel
                     App.Video.VideoEnded -= OnVideoEnded;
                 }
 
-                if (App.LockCard != null)
-                    App.LockCard.LockCardCompleted -= OnLockCardCompleted;
+                // LockCardCompleted is owned by BarkService now (no self-subscription to remove).
 
                 // Unsubscribe from game events
                 if (App.BubbleCount != null)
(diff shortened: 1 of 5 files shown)


CodeBambi:

PR #35, opened 2026-06-05
feat(bark): PR3 — easter eggs (mute swap, note clip, logo wiring, egg conditions)

CodeBambi:
## What this is

PR3 of the bark system: the **easter-egg layer** — mute-egg swap, the real-recording note clip + logo wiring, and the remaining egg condition surface.

> **Stacked on #34 (PR2), which is stacked on #33 (PR1).** Base is `feat/bark-service-pr2-speak`. Merge order: #33 → #34 → #35.

> 🚫 **Do not merge this stack live yet.** Per your instruction, nothing merges until your dry-run validation pass **and** real line content replaces the sample `bark_rules.json`. The bundled note recording (`Resources/sounds/note_newyear.wav`) is also still pending — `PlayNoteClip` no-ops gracefully until it ships.

## Included

**Mute egg (Fork F)**
- `PlayGiggleSound` (`AvatarTubeWindow`) bails before the `Task.Run` when `MasterVolume == 0` — no audio attempt.
- `BarkService.Speak()`: an `easter_egg` bark while `MasterVolume == 0` swaps to a silent, text-only bubble via `Giggle(line)` instead of `GigglePriority` + voiceline.

**New Year note clip**
- `AvatarTubeWindow.PlayNoteClip(string clipPath)` → plays a real recorded clip via the `PlayPhraseAudio` NAudio pattern but with **no giggle sound, no dom voice**, an `_isPlayingUninterruptibleClip` latch (`ShowGiggle`/`Giggle`/`GigglePriority` bail while it holds; the clip's own silent bubble renders via `ShowGiggle(..., bypassClipLock: true)`), and **returns whether a real clip started**. No-ops gracefully if the file is missing.
- `MainWindow.ShowEasterEgg()` → **additive** to the existing written-note dialog: once ever, calls `PlayNoteClip(Resources/sounds/note_newyear.wav)` before the modal note (so the voice plays while the note is read) and latches `AppSettings.NewYearNoteReactionSeen` **only when it actually played** — so it still fires the first time the bundled clip exists.

**Egg condition surface**
- Instant-relaunch threshold tightened to **60s** (`BarkState`).
- Marathon: `SessionEngine.ProgressUpdated` now raises `"SessionProgress"` (periodic tick) so `session_elapsed_sec` thresholds (e.g. 3h) evaluate mid-session.
- Prolonged FaceLost: `FaceFound` raises **before** clearing so `face_lost_sec` reads the elapsed duration (distinct very-long threshold from normal attention-lost).
[...]

PR #36, opened 2026-06-05
feat(bark): PR4 — voiced barks (per-variant voiceline audio)

CodeBambi:
## What this is

PR4 of the bark stack: makes barks **actually speak the supplied voicelines** instead of a generic giggle. This is the engine piece that lets the 3 mods' 251 mp3s/each be heard.

> **Stacked on #35 (PR3) → #34 → #33.** Base is `feat/bark-service-pr3-eggs`. Merge order: #33 → #34 → #35 → #36.
> 🚫 Still part of the held stack — nothing merges until your dry-run validation pass + real content lands (PR6).

## Context: what the content looks like
The supplied `New Voicelines` set is 3 mods (`bambi`, `sissy`, `circe`=the Locked/Circe's Lock built-in; `drone` is the pending 4th) × 251 mp3s, each manifest entry `{ id, n, file, display, spoken }`, plus an authoritative `_spoken.md` that fully specifies every bark's `trigger · condition · priority · cooldown · scope · mood · class`. PR4 makes the engine able to *play* that audio; PR5 closes the few trigger/condition gaps the spec needs; PR6 generates the per-mod rule files and drops in the mp3s.

## Included

**Schema**
- `variant_pool` entries are now `BarkVariant { text, audio? }` and parse from **either** a bare string (text-only, back-compatible with the sample manifest) **or** an object `{ text, audio }`. `display`/`file` accepted as aliases for the content-manifest shape. `BarkRule.VariantPool` is now `List<BarkVariant>`.

**Speak path**
- `BarkService.Speak` resolves the chosen variant's audio filename against the **active mod's** `resources/sounds/companion_audio/` folder (embedded fallback) and plays it as the bubble's audio — no giggle sound on top. Priority barks → `GigglePriority`, queued → `Giggle`; both now thread the voiceline + a `barkVoice` flag.

**AvatarTubeWindow**
- `GigglePriority`/`Giggle`/`ShowGiggle` gain `phraseAudioPath`/`barkVoice` plumbing.
- New `PlayBarkVoice(path)`: NAudio playback gated on **MasterVolume only** (not `SubAudioEnabled`, which is the subliminal-whisper toggle), so companion barks follow master/mute. `MasterVolume==0` → silent.

**Loader: unchanged.** Per-mod `bark_rules.json` already resolves from each mod's `InstalledPath`, so the content side (PR6) just drops a manifest + mp3s into each mod's `companion_audio/`.

## Known limitation (flagged)
A bark routed through the `Giggle` **queue** (bubble already busy) plays **text-only** — the speech-queue tuple carries text only. Barks fire idle in practice (global min-gap + chat-suppression), so the immediate path is the common case; widening the queue to carry audio is a future refinement.

## Validate
Sample manifest now mixes a `{text, audio}` object variant and a bare string in one pool to prove both parse; the audio resolves to null gracefully (mp3 not bundled yet) → text-only. With real mp3s in a mod's `companion_audio/`, the chosen variant speaks. `CCP_BARK_DRYRUN=1` still logs without speaking.

## Build
0 errors; no new warnings in the touched files.

### DIFF
--- a/ConditioningControlPanel/AvatarTubeWindow.xaml.cs
+++ b/ConditioningControlPanel/AvatarTubeWindow.xaml.cs
@@ -2498,7 +2498,7 @@ namespace ConditioningControlPanel
         /// Blocked while waiting for AI response or while AI bubble is visible.
         /// Plays giggle sound 1 in 5 times for preset phrases.
         /// </summary>
-        public void Giggle(string text, string? phraseAudioPath = null)
+        public void Giggle(string text, string? phraseAudioPath = null, bool barkVoice = false)
         {
             if (_isPlayingUninterruptibleClip) return; // an uninterruptible clip is speaking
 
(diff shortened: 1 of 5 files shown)


CodeBambi:

PR #38, opened 2026-06-05
feat(bark): PR6 — content pipeline + 3 mod rule-sets (bambi/sissy/circe)

CodeBambi:
## What this is

PR6 — the final piece: wires the supplied voicelines into the bark system for **all three ready mods** (bambi, sissy, circe). Top of the stack.

> **Stacked on #37 → #36 → #35 → #34 → #33.** Base is `feat/bark-service-pr5-gaps`. Merge order #33→…→#38.
> 🚫 Held with the rest of the stack until your validation pass.

## The transform (committed, re-runnable)
`Tools/bark/build_bark_rules.py` joins:
- **RULE_TABLE** — the authored per-bark metadata (trigger / condition / priority / cooldown / scope / class / mood), transcribed once from the `_spoken.md` spec (identical across mods), with the engine's condition-operator suffixes.
- **`<mod>_manifest.json`** — each mod's per-variant `display` (text) + `file` (audio).

→ a per-mod `bark_rules.json`. Re-run after editing a manifest, adding the **drone** mod (just uncomment it in `MODS`), or tweaking the table. `--copy-audio` also stages the mp3s.

- `kw_reinforce` + `kw_offlimits` → merged into one `kw` bark (per your decision).
- `note_reaction` is skipped (it's the logo-egg recorded clip from PR3, not a bark rule).

## Generated (83 rules each — 84 ids, kw-merge collapses 2→1)
| content | → mod | file |
|---|---|---|
| bambi | builtin-bambisleep | `mods/builtin-bambisleep/bark_rules.json` |
| sissy | builtin-sissyhypno | `mods/builtin-sissyhypno/bark_rules.json` |
| circe | **builtin-locked** (Circe's Lock) | `mods/builtin-locked/bark_rules.json` |

Transform ran clean — every one of the 84 manifest ids mapped, no missing variants.

## Loader / resolver
Bambi & Sissy have **no `InstalledPath`**, so bark content resolves from an **embedded per-mod folder** `Resources/sounds/companion_audio/mods/{modId}/` (`BarkRuleLoader.EmbeddedModManifestPath`; `BarkService.ResolveBarkAudio` gains the per-mod audio path) — uniform across all built-ins. The shared base `bark_rules.json` is now `[]` so per-mod files are the sole source (no samples bleeding into every mod); **CCP Default and Drone are bark-silent** until they get content.

## ⚠️ Decision needed: audio delivery (85MB)
[...]

PR #39, opened 2026-06-05
fix(bark): PR7 — hostile-review fixes (mod-switch reload, variant recycle, thread-safety)

CodeBambi:
Stacked on #38. Fixes from a hostile review of the #33–#38 bark stack. **Behavior-only fixes; no manifest/audio changes.** Builds clean (0 errors).

### 🔴 C1 — Mod switch now reloads rules
`ReloadRules()` had **zero callers**, so the rule-set was frozen to whatever mod was active at startup. Switching Bambi→Sissy mid-run kept Bambi's *lines* and — because audio resolves live by `ActiveModId` and all 248 filenames are identical across mods — played them through Sissy's *voice files* (text/voice mismatch). Now wired into the `ModChanged` handler. Reload also clears the id-keyed rotation / cooldown / session-one-shot state, since rule ids are shared across mods (a Bambi fire would otherwise suppress the same-id Sissy line). Persisted lifetime/tier latches are left intact.

### 🔴 C2 — Repeatable barks no longer go permanently silent
`_usedVariants` was only ever added to, never cleared, so once a repeatable rule used all N variants it returned `variants-exhausted` for the rest of the process. With pools of 3–6, the flash/bubble/subliminal barks fell silent within the first minute of a long run. On exhaustion the pool now recycles, reseeded with the last index so the immediate next line isn't a repeat.

### 🟠 M1 — One-shot rules use their whole pool
One-shots always spoke variant 0 (rotation was gated on `Repeatable`), leaving extra authored variants (e.g. `blink_t1..t5`) unreachable. One-shots now pick a random line.

### 🟠 M2 — `BarkState` is thread-safe
Its mutators run on off-thread source events (webcam tracking) while the matcher reads under `BarkService`'s lock — a torn counter / "Collection was modified" hazard. Now internally synchronized: `Interlocked` blink tally, a lock around the mod-switch list / face-lost timer / marathon set, and a volatile phase name.

### Not changed (by-design / your call)
English-only lines, queued-Giggle audio drop, `♡` note placeholder, raw audio in history. The default `CCPDefault` mod still has no bark manifest — fresh installs get barks once a content mod is selected (now without a restart, thanks to C1).

### DIFF
--- a/ConditioningControlPanel/Services/Bark/BarkState.cs
+++ b/ConditioningControlPanel/Services/Bark/BarkState.cs
@@ -12,8 +12,14 @@ namespace ConditioningControlPanel.Services.Bark
     /// </summary>
     public class BarkState
     {
+        // Source events fire on assorted threads (webcam tracking is off the UI thread) while the
+        // matcher reads this state under BarkService's own, different lock. This lock makes the
+        // mutable collections/counters here internally consistent so a concurrent write+read can't
+        // tear a counter or throw "Collection was modified".
+        private readonly object _lock = new();
+
         // --- session engine (MainWindow-owned, attached late) for live elapsed/phase reads ---
-        private SessionEngine? _engine;
+        private volatile SessionEngine? _engine;
 
         public void AttachSessionEngine(SessionEngine engine) => _engine = engine;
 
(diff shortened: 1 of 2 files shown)


CodeBambi:

PR #40, opened 2026-06-06
feat(bark): mute-whispers silences voicelines; don't talk over whispers

CodeBambi:
Stacked on #39. Two companion-audio adjustments requested during live validation.

## 1. Mute Whispers now silences bark voicelines too
`PlayBarkVoice` was gated on `MasterVolume` only — it deliberately ignored `SubAudioEnabled`. But bark voicelines *are* whispers from the user's perspective, and **"Mute Whispers"** (the avatar menu item + companion-tab checkbox) is exactly `SubAudioEnabled` inverted (`MainWindow.xaml.cs:11396`, `MenuItemMuteWhispers_Click`).

Now gated on **both**: when whispers are muted the **bubble still shows, with no voiceline audio** — same behavior as the existing `PlayPhraseAudio` path.

> ⚠️ Implication: `SubAudioEnabled` defaults to **false** (whispers off). So out of the box, voiced barks now render as silent text bubbles until the user enables whispers. This is the requested coupling — flagging it so it's a conscious choice during validation.

## 2. Barks no longer talk over subliminal/flash whisper audio
Added a shared, thread-safe "whisper busy" signal on `AudioService`:
- `MarkWhisperAudio(durationSeconds)` — stamped by the subliminal (`PlayWhisperAudio`) and flash (`PlaySound`) paths with the clip duration (+0.25s tail). Extend-only via an Interlocked CAS loop, so a short clip can't cut a longer one short.
- `IsWhisperAudioPlaying` — read by the bark gate.

`BarkService.EvaluateGate` now suppresses **non-safety / non-guaranteed** barks while a whisper is audible (gate reason `whisper-active`). Safety/panic and guaranteed reactions still bypass and speak.

### Behavior note — suppress vs. defer
This **suppresses** the bark (drops the line, consistent with the existing `safety-active` / `chat-suppressed` / `min-gap` gates) rather than **queuing** it to play after the whisper finishes. Suppression is the low-risk choice and kills the two-voices overlap cleanly; true deferral would need a timer + captured trigger/context. Easy to switch to deferral if that's preferred after the dry-run.

## Files
| File | Change |
|------|--------|
| `Services/AudioService.cs` | `IsWhisperAudioPlaying` / `MarkWhisperAudio` signal |
| `Services/SubliminalService.cs` | stamp whisper duration |
| `Services/FlashService.cs` | stamp flash-audio duration |
| `Services/BarkService.cs` | `whisper-active` gate |
| `AvatarTubeWindow.xaml.cs` | `PlayBarkVoice` gated on `SubAudioEnabled` |

Builds clean (0 errors). Held for the same dry-run validation pass as the rest of the stack — does not merge live.


CodeBambi:
**Scope note — two commits added on top during live validation:**

- `175d97b feat(mindwipe): customizable audio clip selector` — MindWipe work that landed on this branch (separate concern, piled here per request).
- `3777c16 feat(bark): fresh reactions` — anti-stale fix: ordinary low-priority barks were queuing behind an on-screen non-AI bubble (the gate's chat-suppression only saw *AI* bubbles as "busy"), so the queue drained slowly and she'd comment on ~10s-old events. Now the gate drops queue-class barks while she's mid-bubble (`IsSpeaking`), so the next event gets a *fresh* reaction. Preempting barks (priority ≥ 100 / non-Normal) are exempt — they already clear the queue via `GigglePriority`.

This PR is now mixed-concern (whisper coordination + MindWipe + anti-stale). Still held for the dry-run validation pass.

PR #68, opened 2026-07-08
fix: 6.2.11 green bugfix batch (quiz archetype, voice [unk], bubble presets)

CodeBambi:
Low-risk batch fixing four v6.2.11 bug reports. Build clean (0 errors). Not yet play-tested.

### #501 - Custom quiz category emits hardcoded "Sissy Princess" archetypes
`QuizService.StartQuizAsync(def)` collapses a custom category's `EnumCategory == null` to `QuizCategory.Sissy` for internal bookkeeping, so the result-prompt dispatch in `SubmitFinalAnswerAndGetResultAsync` hit the hardcoded Sissy branch and ignored the user's archetype table. Now the built-in Sissy/Bambi branches are gated on `isCustomCategory` being false, so custom categories fall through to the dynamic-archetype branch that reads `_currentCategoryDefinition.Archetypes`.

### #505 - Voice lock cards choke on out-of-vocabulary words ("cockslut", "gooning")
The Vosk small model can't pronounce these words, so it emits `[unk]` (normalized to `"unk"`). Each `[unk]` counted as a full-word substitution; two or more dropped the fuzzy match score below the 0.62 threshold, looping "try again, slower".

Fix scales `WordEditDistance` costs x2 (match=0, real sub/indel=2) and charges an `[unk]` substitution **half a word** (1). Deliberately half-cost rather than a free wildcard, so a fully-mumbled all-`[unk]` transcript still fails the threshold and can't bypass a lock card. The similarity denominator scales x2 to match, so ratios are **identical** to the old behaviour for any phrase with no unknown words.

### #502 / #503 - Trigger ("special") bubbles do nothing / not saved to presets
`Models/Preset.cs` only carried `BubblesEnabled` + `BubblesFrequency`. It dropped every trigger-bubble property, so saving/loading a preset silently reset `BubbleTriggersEnabled` (which defaults **off**) - making the feature appear inert (#502 is a symptom of #503). Added `BubbleTriggersEnabled`, `BubbleTriggerChance`, `BubbleSpeedBoost`, `BubbleTriggerVariants` (deep-copied list), and `BubbleAvatarEggEnabled` to the class body, `ApplyTo`, and `FromSettings`.

## Not included (from same triage, larger scope)
- #499 update-restart no-op (needs external updater helper)
- #506 / #508 overlay z-order family
- #504 multi-monitor Deeper video loop
- #507 always-listening voice commands

https://claude.ai/code/session_01C8K3fDYUsmV96sJszQgW3e

### DIFF
--- a/ConditioningControlPanel/Models/Preset.cs
+++ b/ConditioningControlPanel/Models/Preset.cs
@@ -100,6 +100,13 @@ namespace ConditioningControlPanel.Models
         public bool BubblesEnabled { get; set; } = false;
         public int BubblesFrequency { get; set; } = 5;
 
+        // Trigger Bubbles (ambient bubbles that fire a Chaos effect on pop)
+        public bool BubbleTriggersEnabled { get; set; } = false;
+        public int BubbleTriggerChance { get; set; } = 10;
+        public int BubbleSpeedBoost { get; set; } = 0;
+        public List<string> BubbleTriggerVariants { get; set; } = new() { "flash", "subliminal", "pink", "spiral", "glitch", "htlink", "video" };
+        public bool BubbleAvatarEggEnabled { get; set; } = true;
+
         // Lock Card Settings (Level 35+)
         public bool LockCardEnabled { get; set; } = false;
         public int LockCardFrequency { get; set; } = 2;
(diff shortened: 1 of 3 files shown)


CodeBambi:
Superseded: the 6.2.11 green batch shipped long ago in later releases; branch is conflicting with main. Closing.

PR #158, opened 2026-08-05
Goon Game: desktop voice notes (V key), mic failure recovery, honest P2P privacy copy

CodeBambi:
Three goon voice-note fixes, developed on separate branches and merged here.

## 1. Desktop mic was unreachable (`fix/goon-desktop-mic`)
The desktop seat runs the same web UI in WebView2 and all voice machinery existed since PR#117 — but the mic button lives in the arsenal drawer, which desktop plays shut, and it was the only control without a hotkey.
- **Push-to-talk on `V`**: hold to record, release to send; the drawer auto-opens for the note and shuts after. Button path unchanged when the drawer is open. Wire format / 10s cap / phone interop untouched.
- WebView2 permission fix: the mic grant no longer saves into the profile (`SavesInProfile = false`), and a banked Deny is repaired on attach — previously one stored Deny muted the in-app mic forever.
- `reportMicGate()` now names the drawer instead of staying silent.

## 2. Browser mic stuck after one failure (`fix/goon-mic-stuck`)
`getUserMedia` was awaited with no ceiling; releasing the button (or dismissing the iOS prompt) during the wait parked the recorder in `starting` forever — every later press refused as `busy` and mislabeled "that one did not record".
- 30s acquire ceiling + attempt tokens (a new press supersedes a dead one).
- Every failure path (denied, cancelled, mid-note error, empty blob, send failure, hang) now resets recorder + button; 20s settle guard as backstop.
- `MediaRecorder` now started with a 250ms timeslice — WebKit could deliver zero data for short notes, making a real recording report as "didn't record".
- New accurate `micFailed` string ("the mic did not open — try again") distinct from `sendFailed`.

## 3. Privacy/consent copy was false (`fix/goon-privacy-copy`)
Copy claimed "nothing leaves your machine" / "no server ever hears them". Verified data path: media rides the P2P `goon-media` data channel and cannot cross the relay; **voice notes ride the control lane and on relay fallback pass through `/v2/goon/relay` unencrypted**.
- All 7 misleading strings rewritten: shared content is *sent to your opponent*, encrypted on a direct link, theirs to keep; voice copy admits the relay hop.
- Lobby disclosure bug: the "nothing ready to send yet" hint *replaced* the transfer disclosure, so a player could enable sending without ever seeing it — now it prepends.
- Self-test added: no voice string may ever again claim a server is uninvolved.

## Merge notes
Two conflicts in `micHud.js` (parallel helper blocks; tap path now uses `safeCancel()` + keyed-aware hint), resolved here.

## Verification
- All 14 goon self-test suites green post-merge: hud 1693/1693, voice 227, voice-ui 263, encode 190, core 242, flow 318, net 350, transfer 199, assets 427, hostgate 97, invite 199, report 213. `selftest-exec` has one rotation-dependent flaky check that fails intermittently **on untouched main as well** (verified 3 runs) — unrelated. `selftest-avafx` failures are pre-existing on main.
- `dotnet build`: 0 errors (312 pre-existing warnings).

## Left for a human
- Play-test: two seats, drawer shut (`V`) and open (button); `V` vs Q/E, 1–7, Esc, F11; dead in join-code field; confirm the real WebView2 mic opens (log: "microphone permission allowed (voice notes)").
- One phone pass with a genuinely denied iOS permission prompt (behavior inferred from code).
- Product calls: relay-mailbox retention wording, TURN/metered.ca nuance, whether voice should be P2P-only like media, and a UI glance at the longer voice ack sheet.

https://claude.ai/code/session_01TGrTm1ev9kpvVDrZYsvesj

### DIFF
--- a/ConditioningControlPanel/Resources/web/goon/boot.js
+++ b/ConditioningControlPanel/Resources/web/goon/boot.js
@@ -1427,9 +1427,21 @@ function reportMicGate() {
     // whether this host has a microphone API at all. Both are read defensively —
     // a handle-shaped stub (no service = a no-op mic) answers neither.
     let hudHidden = false;
+    let deskWhy = '';
     let micSupported;
-    try { if (mic && typeof mic.shown === 'function' && typeof mic.available === 'function') hudHidden = mic.available() && !mic.shown(); }
-    catch (_e) { /* leave it false */ }
+    /* WHICH piece of chrome, not just "some chrome". Zen and the arsenal drawer
+     * both take the slot away and they are undone by opposite gestures — and the
+     * drawer is the DESKTOP failure this breadcrumb existed for and could not
+     * name: an in-app seat plays with the drawer shut (every payload is on the
+     * number row) and the mic is a flow child of the panel that shutting it
+     * removes. Read defensively: a handle-shaped stub answers none of this. */
+    try { if (mic && typeof mic.hiddenBy === 'function') deskWhy = String(mic.hiddenBy() || ''); }
+    catch (_e) { /* an older handle simply cannot say */ }
+    if (deskWhy) hudHidden = deskWhy === 'zen';
+    else {
+      try { if (mic && typeof mic.shown === 'function' && typeof mic.available === 'function') hudHidden = mic.available() && !mic.shown(); }
+      catch (_e) { /* leave it false */ }
+    }
     try { if (mic && mic.recorder && typeof mic.recorder.supported === 'function') micSupported = !!mic.recorder.supported(); }
     catch (_e) { /* leave it undefined — the gate is then not reported */ }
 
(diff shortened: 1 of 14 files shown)


CodeBambi:
Voice notes are now **P2P-only**, like media — the relay hop is gone rather than disclosed.

- `core/match.js`: new `linkIsP2P` (transport state `ConnectedP2P`; TURN still counts). `sendVoiceFrame` refuses on anything else and `_handleVoice` drops inbound voice frames that came up the server mailbox (defence in depth against an unpatched/modified peer). New `onLinkChanged` edge.
- `voiceService.js`: the link joins the one `available()` predicate, so the mic is absent on a relayed duel; `sendBlob` refuses with its own reason `'relay'` (never falls back), and `onFrame` drops relayed frames a second time.
- UI: phone mic button hides and the desktop V key answers false (both ride the same `live` bit); a link degrading mid-recording cancels the hold on the ordinary path. The lobby voice row now says the link, ahead of every other sentence — the checkbox stays live because it is also the receive gate.
- Copy (`strings.js`): `voice.lead` and `voice.ack.line` drop the caveat — "straight to them over your encrypted direct link, never through our servers" — plus two honest lines for the relayed state (`lobbyRelay`, `relayOff`). "Theirs to keep / cannot be taken back" unchanged.
- Tests: the assertions that used to demand the copy admit the relay hop now pin the P2P-only promise instead, and the gate is pinned at the engine (four non-P2P states, both doors), at the service, and on the availability edge. selftest-voice 270, selftest-voice-ui 269, selftest-hud 1696; core/flow/net/transfer/exec green (avafx's 12 pre-existing failures unchanged). `dotnet build` 0 errors.

PR #234, opened 2026-08-19
fix(programs): the You -> Programs hard freeze on a hot run (#984 #993 #996 #1001)

CodeBambi:
Fixes the You -> Programs hard freeze reported in ccp-bugs #984, #993, #996, #1001. Introduced by cae9438d5 (v6.8.0).

## Root cause

`MainWindow/MainWindow.ProgramsFx.cs` -> `ApplyProgramRailComet`. The Programs tab's T4 rail comet animates over the rail's measured width, so it must wait for layout. Three faults in that wait lined up, and each one alone was survivable:

1. **The width could never be non-zero.** `RailCometHost` is authored `Visibility="Collapsed"` (`Views/Tabs/ProgramsTabView.xaml:515`), and WPF never measures a collapsed element. The gate read `tab.RailCometHost.ActualWidth` while it was still collapsed, so it was structurally always `0`. The `Visibility.Visible` that would have fixed it sat ~20 lines further down, **after** the gate.
2. **The retry outranked user input.** It re-posted at `DispatcherPriority.Loaded` (6). Input is 5.
3. **The one-shot guard guarded nothing.** The retry callback set `_programCometPending = false` *before* recursing, and there was no attempt cap.

Together that is an unbounded chain of above-Input dispatcher work, entered on every visit to a Programs tab at Ignited heat (>= 0.80, and boss days 7/14/21/28 pin heat to 1.0) with Motion FX at Full.

Why it presented the way it did in the reports:

- **No crash log.** Nothing throws. The render thread keeps compositing, so the window paints normally and simply stops accepting clicks.
- **"It freezes every time I open Programs."** Hiding the tab calls `StopProgramIgnitionLoops`, which re-collapses the host and nulls the FX signature, so the next visit re-arms the whole thing.
- **#993's "closing it only minimizes, and then it's responsive."** WM_CLOSE goes to `MainWindow.WindowChrome.cs:316`, which `Hide()`s rather than closes. That falsifies the `IsVisible` guard inside the retry, the chain stops, and the app comes back.

## The fix: three layers

The comet is cosmetic. Its worst failure has to be "no comet", never "no app", so all three faults are fixed independently:

1. **Measure a measurable host.** `RailCometHost.Visibility = Visible` now happens *before* the width read, and a host still reporting nothing falls back to `tab.ProgramDayRail.ActualWidth` (same travel distance, and the rail is measured).
2. **Demote the retry to `Background` (4).** Below Input (5), above nothing that matters. Layout still runs first (Render/Loaded outrank Background), so the re-ask still sees a measured rail - but a pathological repeat now degrades to a missing comet instead of an unclickable window.
3. **Attempt counter capped at 3.** Replaces the bool. Budget is spent *before* the retry is posted and never cleared inside it. `StopProgramIgnitionLoops` resets it, so a tab hidden mid-retry cannot latch the comet off for the rest of the session.

The decision itself is extracted to `Helpers/ProgramCometGate.cs` as a pure function (`Decide(hostWidth, railWidth, attempts)` -> Run / Retry / GiveUp + width + the counter to store back), so the logic that froze the app is unit-testable instead of living inside a dispatcher callback. Returning the counter rather than mutating it means "clear the guard, then recurse" is not expressible any more.

The healthy path is visually unchanged, and the reduced-motion gate (`MotionFx.AllowAmbientLoops`) is untouched.

## The heartbeat blind spot (second commit)

`Services/Video/VideoDiag.cs` posted its dispatcher heartbeat only at `DispatcherPriority.Normal`, which cannot see this bug's shape. Work looping above Input starves the Normal beat too, so the trace said `dispatcher UNRESPONSIVE` - the same line a UI thread blocked inside a native call produces. Two very different bugs, one indistinguishable log line, which is why four reports of this arrived with nothing to go on.

Now a second beat posts at `DispatcherPriority.Input` each tick:

- Input starved while Normal still runs is impossible by blocking (the queue drains in priority order), so it can only mean something above Input is re-posting itself. That gets its own escalating line: `dispatcher STARVED at INPUT priority for Nms while Normal still runs ... the window will repaint but ignore clicks`, with the last UI mark.
- When both are stalled, the existing `UNRESPONSIVE` line now carries the input age too, so input starving measurably *earlier* than Normal (priority inversion) is distinguishable from the two ages moving together (a blocking call).
- `HangContext.Describe()` reports `uiInputStallMs`, so the hang watchdog carries the distinction into a bug attachment.

Cost is one extra `BeginInvoke` per second; both beat delegates are hoisted out of the loop, so the tick allocates less than it did before.

## Third commit (small, adjacent)

`_programParticles` was the only ambient canvas in the app never handed to `RegisterTabFx`, so the Today card's Skia sim sat outside the tab-switch park/resume governor. One line, matching every other tab. Registration is idempotent and a `Stop()`ped canvas ignores `Resume()`, so `StopProgramIgnitionLoops` still wins.

## Verification

- `dotnet build`: clean, 0 errors (warning count unchanged).
- Full suite: **3629 passed, 5 failed** - exactly the known nested-worktree baseline (`YouLibraryDoorTests.EachLibraryLauncherIsOneRowBoundToOneExistingHandler` x4 + `Phase8RedirectContractTests.PatreonStillRedirectsIntoTheSettingsDoorsAccountSection`). Nothing new.
- New `ProgramCometGateTests` (15 tests, all passing): collapsed-host width falls back to the rail; NaN/Infinity/negative widths never become an animation duration; a sub-floor rail defers rather than smearing; the cap terminates when driven exactly as the call site drives it; every retry *spends* budget instead of clearing it; give-up is stable; a reset counter re-arms.

### Human play-test recipe

1. Enroll in a program (Kept or Takeover).
2. Set `CurrentDay` to a boss day (7 / 14 / 21 / 28) - that pins heat to 1.0, i.e. Ignited.
3. Settings -> Motion FX -> **Full**.
4. Bounce You -> Programs -> You -> Programs repeatedly, ~10 times.

Expected: the tab opens every time, the rail comet runs, and the window stays clickable throughout. Before this change, step 4 wedged the app on the first visit.

Worth also confirming at Reduced motion (comet should not start at all) and on a non-boss early day (Cold/Warm tiers, no comet by design).

DO NOT MERGE until that play-test is done.

Co-Authored-By: Claude Fable 5 <noreply@anthropic.com>

Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01H3mrhKZQj2ehTJUhj2ueon

### DIFF
--- a/ConditioningControlPanel/Helpers/ProgramCometGate.cs
+++ b/ConditioningControlPanel/Helpers/ProgramCometGate.cs
@@ -1,0 +1,101 @@
+using System;
+
+namespace ConditioningControlPanel.Helpers
+{
+    /// <summary>What the rail comet should do on this pass.</summary>
+    public enum ProgramCometAction
+    {
+        /// <summary>A usable travel distance was measured - start (or restart) the run.</summary>
+        Run = 0,
+        /// <summary>Nothing measurable yet. Re-ask later, at a priority BELOW input.</summary>
+        Retry = 1,
+        /// <summary>Out of attempts. No comet this visit; the rest of the tab is untouched.</summary>
+        GiveUp = 2,
+    }
+
+    /// <summary>The decision, plus the width to animate over and the attempt count to store back.</summary>
+    public readonly struct ProgramCometDecision
+    {
+        public ProgramCometDecision(ProgramCometAction action, double width, int attempts)
+        {
+            Action = action;
+            Width = width;
+            Attempts = attempts;
+        }
+
+        public ProgramCometAction Action { get; }
+
+        /// <summary>Travel distance for the run. Meaningful only when <see cref="Action"/> is Run.</summary>
+        public double Width { get; }
+
+        /// <summary>The attempt counter to WRITE BACK to the caller's field after this decision.</summary>
+        public int Attempts { get; }
+    }
+
+    /// <summary>
+    /// The rail comet's "may I run yet?" gate, extracted from the FX code so the thing that once
+    /// hard-froze the app is a pure function with tests instead of a branch inside a dispatcher
+    /// callback.
+    ///
+    /// <para><b>Why this exists.</b> The comet's travel distance IS the rail's measured width, so
+    /// the effect cannot start before layout has run. The original gate read that width from a host
+    /// that is authored <c>Visibility="Collapsed"</c> and only made visible AFTER the gate, so the
+    /// width was structurally always 0; the retry then re-posted itself at
+    /// <c>DispatcherPriority.Loaded</c> (6) and cleared its own one-shot guard before recursing,
+    /// with no attempt cap. Loaded outranks Input (5), so the retry chain starved user input
+    /// forever - a silent hard freeze with no crash log, on every visit to a hot Programs tab
+    /// (ccp-bugs #984, #993, #996, #1001).</para>
+    ///
+    /// <para><b>The contract this encodes.</b> Prefer the host's own width; fall back to the rail's
+    /// when the host has not been measured yet; below <see cref="MinTravelWidth"/> retry at most
+    /// <see cref="MaxAttempts"/> times and then give up quietly. The counter is returned rather than
+    /// mutated so the caller can only ever store a bounded value - "clear the guard, then recurse"
+    /// is not expressible here.</para>
+    /// </summary>
+    public static class ProgramCometGate
+    {
+        /// <summary>
+        /// Shortest run worth animating, in DIPs. The comet head is 130 wide and enters from -140,
+        /// so anything under this is a smear inside its own clip rather than a comet.
+        /// </summary>
+        public const double MinTravelWidth = 60.0;
+
+        /// <summary>
+        /// Hard cap on re-asks. Three passes is more than layout has ever needed; past that the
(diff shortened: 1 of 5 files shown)


CodeBambi:
Play-tested on hardware (3 runs): control build @ 9609a9a33 FROZE on bounce 1 (2/2, one core pegged ~140%, SendMessageTimeout/IsHungAppWindow both blind to it - dispatcher-queue starvation is invisible to sent-message probes); this branch RESPONSIVE 10/10 bounces (25-52% CPU settling to 18%). Bonus run with only the diag commit (comet fix reverted) froze AND the new 'dispatcher STARVED at INPUT priority' line named it in real time with the 3s/10s/30s ladder, while the old Normal-priority beat logged 'dispatcher alive' through the entire wedge.

One correction to the 70b1cf4f4 commit message prose: it says 'the Normal beat is starved too' - it is not (Normal(10) > Loaded(7)); the code is built on the correct premise ('while Normal still runs'), only that sentence is wrong. Cosmetic follow-up candidate: the Normal ladder resetting mid-wedge prints 'dispatcher alive' between STARVED lines, which could mislead a log tail read.

PR #237, opened 2026-08-19
fix(braindrain): the blur was invisible where the driver leaves the capture's alpha byte at 0 (#960, #975)

CodeBambi:
## The report

"Brain drain plays the audio file if it is available but nothing visual even with every slider at 100%" — ccp-bugs #960, re-reported as #975, plus ModernSilver1 and nikkidoll on Discord. Machine-dependent, no GPU-vendor pattern (two AMD users, one works), survived two releases of diagnostics-only commits.

## Why two releases of diagnostics found nothing

Both reporters' 6.8.2 logs are **clean**:

```
Brain Drain starting - route COMPOSITOR (UnifiedOverlayHost setting is ON; ...), intensity 50%, melt true, dualMonitor true, tier "Quality", monitors [\.\DISPLAY1* 2560x1440@0,0 | \.\DISPLAY2 2560x1440@-2560,0]
Brain Drain started on compositor layer, intensity 50%, melt true
```

No `did NOT arm`, no no-frames warning from either watchdog. Every check passed because every check genuinely was passing: `StretchBlt` returned true, slots built, `FramesPublished` climbed, the layer drew its frame every tick. The frame was **fully transparent**.

## Root cause

`BrainDrainCapturePump` StretchBlts the desktop into a 32bpp `BI_RGB` DIB section. GDI defines the fourth byte of such a pixel as **unused** — what a device-to-DIB blt leaves there is up to the display driver, and it is `0xFF` on some machines and `0x00` on others.

The pump wrapped those bits as `Bgra8888 + SKAlphaType.Opaque`, assuming the declared alpha type makes Skia force alpha to 1. **It does not.** Verified against the SkiaSharp 2.88.8 we ship:

| source alpha byte | wrap | result |
|---|---|---|
| 255 | `Bgra8888` + `Opaque` | correct |
| **0** | `Bgra8888` + `Opaque` | **fully transparent** |
| 0 | `Rgb888x` + swap | correct |

So on a machine whose driver leaves `0x00`, the entire effect composited at alpha 0 while reporting success everywhere. Audio unaffected — it is a separate service.

The **legacy per-screen-window route never had this bug**: it blts into a DDB via `CreateCompatibleBitmap` and hands it to `CreateBitmapSourceFromHBitmap`, which yields a `Bgr32` BitmapSource with no alpha channel to get wrong. That is why "turn the unified overlay renderer off" was a working workaround, and why every compositor-side diagnostic added so far reported health.

## The fix

Stop reading that byte. `Rgb888x` is Skia's "32bpp, fourth byte is padding, opaque **by construction**" colour type, so the driver's leftovers cannot reach the output. `Rgb888x` is RGBX in memory while GDI wrote BGRX, so a shared colour matrix on the draw paint puts R and B back.

A channel swap is per-pixel and linear, so it commutes with the gaussian and with the melt displacement's resample — riding on the paint costs one pipeline stage and no extra offscreen. It also has no failure mode in the melt chain: forcing the alpha inside an image filter instead (the obvious alternative) silently degrades to opaque **black** through the displacement node, which is why this landed on the wrap.

## Telemetry, so the next report is readable

- **One Information line per pump run** naming what the driver actually handed us: the raw alpha byte (mean/min/max) and the frame's mean luminance. Between them they separate the three ways a "started, publishing, nothing on screen" run can happen:
  - alpha 0 → this bug,
  - luma 0 with good alpha → the desktop read back **black** (hardware overlay plane / HDR / protected content), which draws a dark scrim rather than nothing,
  - both healthy → the pixels are fine and the problem is downstream.
- A **Warning** on each of those two conditions, naming them.
- **Per-monitor DPI** on the existing route line. Mixed-DPI is the one topology where "the overlay is up but invisible" could be a placement bug rather than a capture bug, and a report could not tell them apart.

## Why the suite missed it

Every synthetic source in `BrainDrainMeltFilterTests` is built with `SKCanvas.Clear`, which writes alpha 255. The field's source is only opaque by driver convention. The new `BrainDrainCaptureAlphaTests` use the field's shape — a buffer whose alpha bytes are **zero** — and cover:

- the trap itself (`Bgra8888` + `Opaque` with alpha 0 draws nothing), pinned so a Skia bump cannot quietly change it,
- the production wrap surviving alpha 0 on the plain **and** melt paths,
- a regression guard that machines which always worked render identically,
- the diagnostic sampler, including telling a black capture apart from an invisible one.

## Verification

- `dotnet build` clean.
- Full suite: **3636 passed, 5 failed** — exactly the known nested-worktree baseline (`YouLibraryDoorTests` ×4 + `Phase8RedirectContractTests.PatreonStillRedirectsIntoTheSettingsDoorsAccountSection`).
- All 23 brain-drain tests (7 new + existing melt/hand-off) pass.

Not merged — wants a play-test, and ideally a build to Nardda (BUG-MMDM2LXEL9) and the AU reporter, whose logs will now say outright which condition their machine is in.

https://claude.ai/code/session_01H3mrhKZQj2ehTJUhj2ueon

### DIFF
--- a/ConditioningControlPanel/Services/Compositor/BrainDrainCapturePump.cs
+++ b/ConditioningControlPanel/Services/Compositor/BrainDrainCapturePump.cs
@@ -113,6 +113,52 @@ internal sealed class LatestFrameSlot<T> where T : class, IDisposable
 /// </summary>
 internal sealed class BrainDrainCapturePump
 {
+    // ================================================================================
+    //  #960 / #975 - THE ALPHA BYTE. Read this before touching the wrap below.
+    // ================================================================================
+    //
+    // The staging surface is a 32bpp BI_RGB DIB section and the capture is a GDI StretchBlt off
+    // the desktop DC. GDI defines the fourth byte of a 32bpp BI_RGB pixel as UNUSED: it is under
+    // no obligation to write anything into it, and what a device-to-DIB blt actually leaves there
+    // is up to the display driver and the blit path. Most machines leave 0xFF (the DWM's own
+    // surface carries it) and every dev box we own is one of them. Some machines leave 0x00.
+    //
+    // This code used to wrap those bits as Bgra8888 + SKAlphaType.Opaque and assume Skia would
+    // force the alpha to 1 because the type says the image is opaque. IT DOES NOT (verified
+    // against SkiaSharp 2.88.8): a Bgra8888 buffer whose alpha bytes are 0 draws as FULLY
+    // TRANSPARENT no matter what the declared alpha type says. On a machine whose driver leaves
+    // the byte at 0 the whole feature therefore rendered NOTHING while every single check passed
+    // - StretchBlt returned true, frames published, FramesPublished went up, both first-frame
+    // watchdogs stayed quiet, the audio half ran normally. That is #960/#975 exactly: "the audio
+    // plays but there is no visual", machine-dependent, no GPU-vendor pattern, nothing in the log.
+    //
+    // THE FIX: stop reading that byte at all. Rgb888x is Skia's "32bpp, fourth byte is padding,
+    // the image is opaque BY CONSTRUCTION" type, so the driver's leftovers cannot reach the
+    // output. Rgb888x is R,G,B,X in memory while GDI wrote B,G,R,X, so <see cref="SwapRedBlue"/>
+    // puts the channels back - it rides on the draw paint, which costs one extra pipeline stage
+    // and no extra offscreen. A channel swap is per-pixel and linear, so it commutes with both
+    // the gaussian and the displacement resample: applying it after the filter chain is exactly
+    // equivalent to applying it to the source.
+    //
+    // The legacy per-screen-window route never had this bug: it blts into a DDB via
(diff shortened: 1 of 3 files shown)


CodeBambi:
In-house play-test result: the owner's dev machine turns out to BE an alpha-0 machine - the new telemetry logged 'raw GDI alpha byte mean 0.0 (min 0, max 0)' with the #960/#975 warning firing. A/B against the pre-fix build on this box: pre-fix composites ADDITIVELY (red patch 157.6 -> 231.1 brightness, magenta seam where red meets blue = src + dst*(1-0) with alpha 0 premultiplied), which reads as 'nothing' on a dark desktop - matching every field report. Post-fix: true blur preserving mean brightness (249.9/0.0/0.0 on pure red, 0.0/0.0/249.8 on pure blue, zero channel bleed, 3 stable frames), melt animating correctly, 20/20 Skia-level probe exact across alpha {0,255} x melt {on,off}. Note for the PR body: 'dev boxes are all 0xFF' is wrong - this fix is verifiable in-house, and the pre-fix additive artifact may explain past internal 'washed out / does nothing' impressions.

PR #373, opened 2026-08-28
feat(arcademy): wallet banks on the server, shared across devices

CodeBambi:
## What this is

The desktop half of the shared Arcademy wallet. Today every platform mints and spends locally, so
the balance on the PC and the balance on the phone are two different piles of tickets. This puts the
wallet on the account: the server owns it, the desktop keeps its local copy as a cache of the last
answer, and a second device sees the same balance on its next launch.

Implements the "Client behaviour" section of the wallet contract. The server lane is
`CC-Labs-llc/CCP-Server` branch `feat/arcademy-wallet`; the web and mobile lanes are being built to
the same contract in parallel.

## Why it is shaped like this

`Services/Arcademy/ArcademyWalletSyncService.cs` is deliberately the punch-card mirror's twin: same
proxy, same `X-Auth-Token` + `unified_id` door, same fire-and-forget `Run` wrapper, same generation
counter so a late reply cannot write into the next window, same quiet logging (`[ArcademyWallet]`).
Nothing here blocks the page and nothing here puts a dialog in front of anybody.

Two rules held the design:

**Never subtract locally.** No failure path ever takes tickets away. The server is the only thing
that can spend.

**Never double-pay.** Every graded finish carries a client-minted `mintId`, the server keeps those
for 30 days, and a replay of the same frame returns the stored result rather than paying again.

## Offline queue

- Signed out: today's local-authority path, byte for byte. Nothing was touched.
- Server answers: adopt `economy.wallet`, post `payout-result` from the server's numbers.
- 5xx / 409 / timeout / network / 401: mint LOCALLY so the debrief still shows a number, then park
  the frame in `pendingMints` with `queued:true` and the SAME `mintId`. The next launch, or the next
  mint that lands, replays the queue oldest first and the server's answer replaces the local wallet.
  The local mint was only ever a preview.
- 403 tier: mint locally, park NOTHING, log once per launch. A queue of frames that can never land
  is just a leak.
- One extra guard the contract does not name: `Park` refuses to queue anything before this machine
  has imported. Until the import runs the local wallet is still the record, and the import itself
  carries that night's money up. Parking as well would bank it twice.

## Import rule

On launch, with an account behind it, GET the wallet. If this install has never imported and either
the server has a wallet or the local one has earnings, POST the import once (server merges, keyed on
a per-install `walletDeviceId`), then adopt. From then on local is a cache. If neither side has
anything yet, nothing is sent. `walletImported` and `walletDeviceId` live in host-owned meta, not in
the wallet object.

`init.economy.payday` shows the server's draw when the launch pull got through, and the local
`PickPayday` otherwise.

## Two reasons the Prize Counter had no words for

`Resources/web/arcademy/shell/prizecounter.js`'s REFUSALS map had no row for `offline` or `busy`, so
both fell to the unknown shrug line. Once the wallet is on the account those are the two refusals a
purely local till could never produce, and neither is anything the player did:

- `offline` - the counter cannot reach the bank right now. Nothing was charged.
- `busy` - somebody is already at the drawer (one of the player's OWN other devices, mid-write).

Added to the map, to `core/lexicon.js`, and to the host lexicon in `ArcademyHostService` alongside
`prize_poor` / `prize_locked`. Web and mobile vendor this page byte for byte, so the rows land on
every platform from this one branch. Nothing else in `Resources/web` was touched.

Failure never zeroes the chip either: an unanswered Buy echoes the current local wallet back with
`ok:false`, never a null one.

## What was verified

- `dotnet build`: 0 errors, 344 warnings, all pre-existing (CA1416 and friends). No warning in any
  Arcademy file.
- An offline harness in the scratchpad compiles the REAL `ArcademyWalletSyncService`,
  `ArcademyMetaStore`, `ArcademyEconomy` and `ArcademyPunchCards` (one edit: the proxy base URL
  points at a local `HttpListener`) and drives them against a fake proxy. 18 checks over 6 scenarios,
  all passing:
  - launch adopts the account wallet whole (1234t / 9k), stamps `walletImported`, sends this
    install's `deviceId` in the import body, and takes the server's payday (sort x5)
  - a 500 comes back as "queue", the frame is parked under the same `mintId`, and the frame carries
    the wire's field names
  - a 200 with an `economy` block is banked and 184t / 2k is adopted
  - the replay drains the queue, re-sending the original `mintId` with `queued:true`
  - a 403 comes back Refused and parks nothing
  - before the import, nothing is parked at all
- Field names read against the live server branch: GET wallet (`exists`, `wallet`,
  `payday{gameKey,mult}`, `catalogWave`, `serverDay`), import body and its `{imported, wallet,
  clamps}` answer, class-ended's nested `economy.wallet`, prize-buy's 409 `{reason:'busy'}` and its
  200 `{ok, sku, reason, wallet}`. All match.
- Not verified by a play-test: this has not been run against the real proxy, because the server
  branch is not deployed.

## REPORT: which tier the desktop lets into the Arcademy

The door is `TierGate.DemandLab` -> `App.Patreon.HasLabAccess`, which is true when ANY of these
hold:

- Patreon `CurrentTier >= Level2`
- Patreon `IsWhitelisted`
- SubscribeStar tier >= 2
- SubscribeStar `IsWhitelisted`
- `HasCachedLabAccess` - a 14 day offline grace on the last good validation

The server's class-ended gate is strictly tier >= 2 (whitelist included, same arithmetic minus the
grace). So the two are the same set EXCEPT for the grace window: an account whose subscription
lapsed, or whose validation cannot reach the server, can still play locally for up to 14 days and
will 403 on the mint. That case is handled - local mint, nothing queued, one log line - but it is
worth an owner decision: the player keeps earning on that machine and those tickets never reach the
account.

## Open question for the server lane

`/v2/arcademy/class-ended` still resolves through `resolveBearerCaller`, which wants an
`Authorization: Bearer` header. The desktop sends none - it carries `X-Auth-Token` + `unified_id`,
the pair `resolveAccessCaller` already handles for `/access` and the pair the new wallet routes
accept. Until that route moves to `resolveAccessCaller`, every desktop mint 401s. The client treats
a 401 as "queue it, do not lose the money" with its own once-per-launch warning, so nothing breaks
today; the tickets simply stay parked until the door opens. Raised with the server lane.

Do not merge this ahead of the server branch. Signed-out players are unaffected either way.

### DIFF
--- a/ConditioningControlPanel/Resources/web/arcademy/core/lexicon.js
+++ b/ConditioningControlPanel/Resources/web/arcademy/core/lexicon.js
@@ -600,6 +600,8 @@ export const DEFAULT_LEXICON = Object.freeze({
   prize_full: 'Your pockets are full of those. Use one first.',
   prize_locked_msg: 'That one stays in the case for now.',
   prize_unknown: 'The counter does not know that one. Odd.',
+  prize_offline: 'The counter cannot reach the bank right now. Nothing was charged.',
+  prize_busy: 'Somebody is already at the drawer. Give it a second and ask again.',
   prize_quiet: 'The counter went quiet on that one. Try again in a moment.',
   prize_empty: 'Shelf is bare tonight. Come back when the truck has been.',
   prize_payday_label: 'Hot room tonight',
--- a/ConditioningControlPanel/Resources/web/arcademy/shell/prizecounter.js
+++ b/ConditioningControlPanel/Resources/web/arcademy/shell/prizecounter.js
@@ -62,6 +62,14 @@ export const REFUSALS = Object.freeze({
   full: ['prize_full', 'Your pockets are full of those. Use one first.'],
   locked: ['prize_locked_msg', 'That one stays in the case for now.'],
   unknown: ['prize_unknown', 'The counter does not know that one. Odd.'],
+  /* THE SHARED WALLET (2026-08-28). The wallet is on the account now, so two
+   * refusals arrive that no local till could ever have produced: `offline` when
+   * the host could not reach the bank, and `busy` when one of your OWN other
+   * devices is mid-write on the same wallet. Neither is a refusal the player did
+   * anything to earn, and in both cases nothing was charged - which is the half
+   * worth saying out loud. Without these rows both fell to the unknown shrug. */
+  offline: ['prize_offline', 'The counter cannot reach the bank right now. Nothing was charged.'],
+  busy: ['prize_busy', 'Somebody is already at the drawer. Give it a second and ask again.'],
 });
 
 /** The glyph each sku wears on its little drawn box. Text, not art: a sku the
--- a/ConditioningControlPanel/Services/Arcademy/ArcademyHostService.cs
+++ b/ConditioningControlPanel/Services/Arcademy/ArcademyHostService.cs
@@ -206,6 +206,12 @@ internal static class ArcademyHostService
             // exactly the same, on the cards this machine already holds.
             ArcademySyncService.Attach(_meta, OnMirrorCardsChanged);
 
+            // THE SHARED WALLET (wallet contract, "Client behaviour"). Same posture, one shelf
+            // over: fetch the ACCOUNT's wallet, carry this machine's up if it never has been, and
+            // drain anything a night offline left unpaid. Signed out, this does nothing at all and
+            // the money path below stays exactly the local-authority one it has always been.
+            ArcademyWalletSyncService.Attach(_meta, OnWalletBanked);
+
             // CAMPUS PRESENCE (PRESENCE.md §3). Two independent halves behind one Attach: the
             // EMITTER announces `campus_enter` (only if the player has opted in - the rung is read
             // inside), and the SNAPSHOT PUSHER starts polling the public feed and handing it to
@@ -753,7 +759,12 @@ internal static class ArcademyHostService
     {
         try
         {
-            var payday = ArcademyEconomy.PickPayday(
+            // THE SERVER'S DRAW WINS WHEN THERE IS ONE. Both sides run the same seeded pick over
+            // the same roster, so they normally agree to the letter; where they can differ is a
+            // roster that has not finished mirroring, and on that night the account's answer is the
+            // one every OTHER device is also showing. The local pick is the fallback, and it is
+            // what a signed-out player (and an early `init` that beat the reply home) gets.
+            var payday = ArcademyWalletSyncService.ServerPayday ?? ArcademyEconomy.PickPayday(
                 DateTime.UtcNow.ToString("yyyy-MM-dd", CultureInfo.InvariantCulture),
                 _meta?.EnrolledGameKeys());
             var (extra, honors) = _meta?.LeverUnlocks() ?? (false, false);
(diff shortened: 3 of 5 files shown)


CodeBambi:
**Wire alignment pass, desktop lane.** Server read off CCP-Server `feat/arcademy-wallet` (routes + `arcademy-routes.js` class-ended + test section 17), not just the doc. Three commits on this branch.

**The agreed behaviour change (`339e2ba8`)**

- A **403 TIER refusal on class-ended no longer drops the frame.** It parks under the SAME `mintId` with `queued:true` and replays on the NEXT LAUNCH only, one log line per launch, no retry storm inside the session. Why it overrides the server doc: the desktop grants a 14-day cached-entitlement grace the proxy does not, and the once-per-device import can never carry tickets earned AFTER it ran. The queue can, and the server pays a queued frame even when the stamp is outside the attendance wall (`<game>:queued_stamp_walled`).
- Scoped to `code == 'arcademy_locked'` or a `min_tier` present. A 403 `arcademy_identity_unknown` is a different refusal (no account to bank into) and keeps its old ending: local mint, nothing parked.
- Same commit: a 400 `local_day_out_of_range` off an **unqueued** frame now parks too. class-ended keeps two day walls and the tight one is what an unqueued frame meets; the same frame replayed as `queued` gets the 14-day window, so that refusal is one a replay can clear. Every other 400 stays dropped.
- The bank-open latch is untouched: 401 / 404 / a 200 with no `economy` block still fall to the local path.

**Mismatches found and fixed**

- `ddc719c1` import: `adopted` was never read, so an adopt-whole was logged as "merged".
- `ddc719c1` prize-buy: no `localDay` was sent, so the receipt and inventory `at` took the proxy's UTC day - tomorrow's date on a prize bought tonight, west of UTC.
- `ddc719c1` class-ended: `clamps` was never logged, so `queued_stamp_walled` was invisible in a support log.
- `2ad1e56c` `deviceId` was only length-checked, not shape-checked. An id of the right length in the wrong alphabet would fail the route's `^[A-Za-z0-9_-]{8,64}$` on every launch for ever, with the import the one call that cannot be skipped and retried.

**Re-verified clean, no change needed:** door 3 on all four routes (`X-Auth-Token` + `unified_id`, query on the GET, body on every POST); prize-buy's `wallet` read TOP-LEVEL; the 409 split (prize-buy `reason:'busy'`, class-ended `wallet_busy` vs `cards_busy` with no reason, both parked and both idempotent on replay); `minted:0` cannot be misread because nothing here reads `minted`; `mintId`/`deviceId` are `Guid.ToString("N")`, 32 lowercase hex; `lever` is always one of the three (ClampLever); `streak` clamped 0..3650; `zen`/`lateSlipUsed`/`queued` are real JSON booleans; every game key matches `^[a-z][a-z0-9_]{0,31}$`.

**Proof:** the offline HttpListener harness was extended - a 403 tier body parks under the same `mintId`, one attempt only inside the session, and the next launch replays it `queued:true` and banks it through a `minted:0` + `queued_stamp_walled` answer; a 403 `identity_unknown` still parks nothing; the two kinds of 400 are told apart. 25/25 checks pass. `dotnet build`: **0 errors, 344 warnings** (all pre-existing, none in Arcademy files).

PR #409, opened 2026-08-30
fix(ask-emi): the book takes a narrow width rather than parking on top of her

CodeBambi:
## The book stops parking on top of her

The clamp at the end of the book's placement dragged it back over her whenever neither side of her
had room:

```csharp
Left = Math.Max(workL, Math.Min(workL + workW - w, left));
```

Her body, her gear, her `?` and anything in her hand, all under the panel. It needs a work area under
roughly `bodyWidth + 750`, so it never showed on the machines the book was written on and would have
shown on somebody's 1280 laptop with her scaled up. Found while answering a question about whether
the book could collide with #407's prop plate - it cannot, but this can.

### Two widths

If the full 364 fits on either side of her, **nothing changes**. If it does not, the book goes narrow:
270 DIP, the same chrome, the same rail, the demo stage at 2x, and it stays *beside* her. A smaller
book that can be read beats a big one parked over the thing it is explaining.

Only when even the narrow book has nowhere to go does it overlap her - and then it says so in the log,
because every visual symptom of that (panel over her body, her chips unreachable) looks like a z-order
bug and is not one.

Two widths and not a continuous one on purpose: the cards are written to a measured column, and a book
that took whatever width was going would give every desk a different one.

### Why the geometry moved out of the window

`Services/EmiDesk/EmiBookLayout.cs` is now pure arithmetic. That is the whole point - the bug was
unreachable from a test while it lived inside a `Window`, and it is exactly the class of bug that only
appears at a screen size nobody on the project owns.

`EmiBookLayoutTests` (20) sweeps her a DIP at a time across 1920, 1600, 1366, 1280 and 1024 desks and
asserts the book is never on her body without having declared it, never off the screen, always one of
the two widths, and exactly one `BodyGap` off her edge whenever it fits.

### The bullet that was losing its last word

Fell out of the narrow reflow, but it was already true at full width on a short desk. The nudge list
scrolls with `VerticalScrollBarVisibility="Hidden"`, so a card taller than its room silently loses the
end of its last bullet. At 196 the LOCKDOWN card rendered

> five clicks on the timer, under a second apart, then type **let me**

and dropped `out`. That is the escape phrase, and the one line in the book somebody reads in a hurry.

A small pink chevron now shows while there is copy below and goes away once you reach it. Drawn rather
than a real scrollbar, for the reason the original comment gave: a scrollbar is the one piece of chrome
on this window that Windows would draw instead of us.

### Rig

`--shoot-book <dir> --narrow` forces the narrow book, so the reflow can be reviewed on a desk that has
room. Nothing in a normal launch sets it.

## Tests

4842 pass. The one red test is `EmiAliveTests.NoFidgetEverRepeatsItselfAndAllThreeGetUsed`, which is
red on `main` already and is being fixed on `feat/emi-idle-rotation` - this branch does not touch the
fidget pool, `EmiProps` or the desk window.

## Verified how

The narrow book was rendered offscreen at 270 and read card by card: the rail keeps its icons, the 2x
stage fits the column exactly, and the densest card (LOCKDOWN, eight glyphs in the rail) still lands
its catch strip and its button without clipping.

**The chevron itself has not been shot yet** - a peer session's build holds the single-instance mutex
and the offscreen rig cannot start behind it. Its logic is trivial and covered by inspection, but it
has not been seen. I will shoot it and post the frame here as soon as the app is free.

https://claude.ai/code/session_012nMSMQZ2ieD7Hp9enNq63J

### DIFF
--- a/ConditioningControlPanel/App.xaml.cs
+++ b/ConditioningControlPanel/App.xaml.cs
@@ -26,6 +26,15 @@ namespace ConditioningControlPanel
     public partial class App : Application
     {
         /// <summary>
+        /// True when this launch is an offscreen screenshot rig (<c>--shoot-book</c>,
+        /// <c>--shoot-doors</c>, <c>--possession-preview</c>) and there is nobody to answer a dialog.
+        ///
+        /// <para>Read it to skip a MODAL question, never to skip the thing being tested. A rig exists
+        /// to photograph the shipped behaviour, so anything that draws must still draw.</para>
+        /// </summary>
+        public static bool IsUnattendedRig { get; private set; }
+
+        /// <summary>
         /// Custom entry point. Originally added for Velopack's update hooks; kept after
         /// Velopack removal (v5.8.4) so we still control startup ordering explicitly.
         /// </summary>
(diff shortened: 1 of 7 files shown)


CodeBambi:
Shot it, and the first version of the cue was wrong in a way only a picture would have shown.

**The chevron sat on the words.** Floated over the bottom right of the copy, it landed on the very sentence this change exists to protect - LOCKDOWN rendered `then type let m [v]`. A cue that hides the words it points at is worse than no cue, because the reader cannot even tell something is missing. It now takes its own band under the bullets. A row cannot oscillate: showing it shrinks the viewport, which can only add overflow, never remove it. It sits inside the bullets' own `*` row, so the catch strip and the button do not move when it appears.

**And the rig was never unattended.** `--shoot-book` drives the shipped app and summons her the way a user does - that is the point of it - but `MaybeAskAboutMuting` opens a modal dialog on that path, the summon is abandoned while it is up, and the rig waits ten seconds for a companion who is never coming and writes `she never came out`.

It failed silently and intermittently: the prompt is once per session and only appears when something else is talking, so the same command wrote 132 shots one minute and zero the next. **The run that passed passed because a person happened to click the dialog while it was going**, which is the worst possible way for a rig to succeed. A rig launch now declares itself once in `App.OnStartup` and the prompt takes the answer a dismiss would have given. Nothing that draws is suppressed.

That one also covers `--shoot-doors` and `--possession-preview`, which had the same latent hang.

## Verified

Both books shot end to end, 132 frames each, with nobody at the keyboard:

- **270 (narrow)** - LOCKDOWN keeps its text and the chevron sits under it, clear of the words. The rail keeps all eight DEEPER icons at full size, the 2x stage fits the column exactly, catch strip and button land without clipping.
- **364 (wide)** - the same card renders `then type let me out` in full, with no cue at all, exactly as it does on `main` today.

4842 pass. The one red test is the fidget pool one that is already red on `main` and fixed on `feat/emi-idle-rotation`.

PR #618, opened 2026-09-03
feat(core): CoreDispatch, the UI-thread seam, and its two consumers

ObviouslyNotMich:
First layer of the service-wiring project, and the first thing that project needs.

## What
`CoreDispatch` is a two-delegate seam matching `CoreMods`: hop to the head's UI thread if there is one, run in place if there is not, bounded so a wedged head cannot stall the engine. The WPF head seeds it; `SettingsService.SerializeCurrentSnapshot` and `ModService.RunOnUi` now go through it.

Those two were the only places the settings and mod layers named `System.Windows`. Behaviour is unchanged, fallbacks included: running in place with no UI thread is exactly what both already did under the test host and the Linux smoke runner.

## Why this and not the settings move
The obvious first layer was `git mv` of `AppSettings` into Core. It measured clean — no WPF usings, and no WPF-typed properties — so I tried it, and it does not work. Two findings:

- **The 8,164-line settings model calls back into the WPF locator.** Eleven references across `App.Bark`, `App.Mods`, `App.Brain`, `App.ResolveScreens`, `App.OnStartup` and `App.EnsureInstallDateRecorded`. Each needs its own seam before the model can move.
- **Its dependency closure bottoms out in `System.Windows.Rect`.** The model needs the mod manifest, which needs the art-framing table, which types a shipped viewbox as a bare `Rect`. That is the trap `CLAUDE.md` names. It is fixable — the rect is pure data and there is exactly one live conversion point in the head — but it is a layer of its own, not a rider on a move.

So the move is deferred rather than forced, and this layer lands the part that is ready and already has consumers.

## Deferred, named so nobody re-derives it
`VideoService` (167 WPF references) is not portable and is out of scope for this project. `TutorialService` (11) is a judgement call. `BuiltInMods`, `HapticMixer` and the two auth stores are WPF-free and move cleanly when their turn comes.

## Verified
Solution builds, Core guards pass, the engine executes on Linux, `--smoke` passes, 186/186 views render.

### DIFF
--- a/CCP.Core/CoreDispatch.cs
+++ b/CCP.Core/CoreDispatch.cs
@@ -1,0 +1,83 @@
+using System;
+
+namespace ConditioningControlPanel
+{
+    /// <summary>
+    /// The UI-thread seam. Engine code that must run a fragment on the head's UI thread asks here
+    /// rather than through <c>System.Windows.Application.Current.Dispatcher</c>, which lives on a
+    /// WPF type and therefore cannot exist in Core.
+    ///
+    /// <para>Two call sites need this today and they need the same thing: a bounded hop to the UI
+    /// thread with a fall back to running in place. <c>SettingsService</c> retries a serialize
+    /// there when it races a UI-thread edit, and <c>ModService.RunOnUi</c> marshals a collection
+    /// mutation. Both already treat "no UI thread" as normal, because the test host and the Linux
+    /// smoke runner have none.</para>
+    ///
+    /// <para>Deliberately two delegates and no interface, matching <see cref="CoreMods"/>: every
+    /// head seeds them the same way, and an interface with one implementation is speculation.
+    /// Promote when a second consumer needs something these cannot express.</para>
+    ///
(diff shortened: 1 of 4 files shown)


CodeBambi (changes requested):
Seam looks right and this is the one the agreement named first. Please retarget it to main, keep the `seam` label, and one fix:

**Shutdown semantics changed.** The old `ModService.RunOnUi` dropped the action when the dispatcher had started shutting down. The WPF `PostProvider` in the App static ctor now runs it in place on the calling thread:

```csharp
if (d is null || d.HasShutdownStarted) { action(); return; }
```

CoreDispatch's own doc says a shutting-down head "should seed a provider that drops the action". Keep run-in-place for the null dispatcher case only and return early when shutdown has started. We have had real crashes from UI work after shutdown starts, which is why the old code dropped it.

Minor: `SerializeCurrentSnapshot` now retries once on the same thread when `CheckAccess()` is already true (old code rethrew immediately), and the "raced a UI-thread edit, retrying" debug line got lost. Not blocking.

PR #634, opened 2026-09-03
feat: one shared /Assets root, linked by every head

ObviouslyNotMich:
The art moves out of the WPF head to a sibling `/Assets` root, and every head links what it needs from there.

## Why the move, not just the packaging
Linking the Avalonia head's art out of `ConditioningControlPanel/Resources` makes the head we are retiring the source of truth, and CCP.VR and a future mobile head would have had to reach into it too. `/Assets` is a sibling of the projects and belongs to none of them.

## What stays in the WPF head, deliberately
`ConditioningControlPanel/Resources/Theme/`. Those are WPF resource dictionaries that `App.xaml` merges at startup: code, not art, and SDK-globbed as `Page` only while inside the project directory. The first cut of this layer moved them and the build still passed, because the loss only shows at runtime on Windows. They are back where the head compiles them. The Avalonia head ported that palette key for key and links nothing from it.

## Logical paths preserved, proven rather than assumed
`Link=` keeps every `pack://` URI in the WPF head and every `avares://` URI in the Avalonia head unchanged, so no view needed editing. Three traps in getting that right, each fixed here and recorded in `CLAUDE.md`'s neighbour notes:
- `%(RecursiveDir)` is empty for a single-star glob, so a bare `Link="Resources\%(RecursiveDir)…"` flattened every emote folder into one and `comehere.gif` collided. Every glob's Link now names its folder.
- Items carrying a `Condition="Exists(…)"` need the condition repointed and a Link of their own, or their key collapses to the bare filename (`emotes.json`, four times).
- The keys were then verified by searching the built assembly for the UTF-16 strings `resources/circe_emotes_p2/emotes.json`, `resources/features/awareness.png` and `resources/theme/colors.baml`, all present.

Runtime paths still resolve because `Content` items copy to output under the same `Resources/` path. The publish targets, the installer icon, the stream-overlay script and the awareness-sound tool are repointed. Sixteen Windows tests that built source-tree paths into the art now build them into `/Assets`; the ones that use the WPF `Resources` folder only as a repo-root marker are untouched, since that folder still exists.

## What the Avalonia head takes
Only what its views name, measured from the 113 `pack://` URIs in the WPF source: **67.6 MB against the WPF head's 257.5 MB.** Twemoji (4,009 files, 19 MB) is excluded entirely, since Avalonia renders colour emoji natively. `achievements/` and `skills/` are listed file by file. `Assets/README.md` documents the item shapes for a new head.

## Proof
`AwarenessTabView`'s hero plate is restored so the path is exercised end to end; its render is byte-identical before and after the fix.

## Verified
WPF head, Avalonia head and the Windows test project all build from scratch (`--no-incremental`), Core guards pass, `--smoke` passes on both heads, 186/186 views render.


CodeBambi (changes requested):
Still held until the day after 6.9.2 is tagged, as agreed. Windows CI is green now so that precondition is met. The mechanical part looks right: the `Link=` rewrite keeps every pack:// URI, installer icon and tests are repointed, Resources/Theme stays. Two things before it is re-requested:

**1. Pack-only audio now ships in the free build.** The relative `ContentPackSoundsExclude` (csproj around lines 479-495) still lists `Resources\sounds\flashes_audio\**\*.mp3`, `Resources\sounds\builtin-bambisleep\**\*` and the rest, but its consumer moved:

```xml
<Content Include="..\Assets\sounds\**\*" Link="Resources\sounds\%(RecursiveDir)%(Filename)%(Extension)"
         Exclude="$(ContentPackSoundsExclude)" ... />
```

MSBuild matches Exclude against the Include path, so nothing is excluded any more. Every tracked file in those folders (118 in flashes_audio, 1,469 in builtin-bambisleep, plus the others) lands in the publish folder and then in Setup.exe. That is paid content-pack audio in the free download. The `Abs` variant around line 499 was updated, the relative one was not. Same treatment, or derive one from the other.

**2. Runtime check.** Nobody has launched the Windows app from this branch. Before re-request: one `dotnet publish`, confirm `publish/Resources/sounds/flashes_audio` does not exist, then one launch plus one content-pack install to confirm the Link mapping at runtime.

Also please move this PR to the end of whatever stays stacked. At position two it blocks the eight PRs behind it for no reason.

PR #635, opened 2026-09-03
feat(avalonia): the shell is the startup window; diagnostics moves behind a button

ObviouslyNotMich:
The head opened its diagnostics `MainWindow` at startup. That was right while the app shell did not exist, and wrong now that it does.

## What changes
- **Startup is `MainShellWindow`.** One line in `App.axaml.cs`.
- **Diagnostics moves behind a small muted `diagnostics` button** at the end of the Settings action row. Deliberately not styled as a setting: it is a developer affordance.
- It opens as a **second window owned by the shell**, not a swap, so the shell keeps its state. A second click focuses the window already open rather than stacking duplicates.
- **The diagnostics window gains a `← Back to the app` button**, which is therefore just a close.

## Two things worth a reviewer's eye
- The Back button **hides itself when there is no owner**. `RenderProof` also hosts single views inside a `MainWindow`, and a dead control sitting in 186 render proofs is noise. It appears only when a shell actually opened the window.
- **`RenderProof` is untouched.** Hosting a view in a `MainWindow` never depended on that window being the startup one, so the render path is unchanged and still proves what it proved before.

## Verified
Solution builds, Core guards pass, `--smoke` passes, 186/186 views render, and the app launches to the shell on KDE Wayland with an empty log.

75 changed lines across 5 files.

### DIFF
--- a/CCP.Avalonia/App.axaml.cs
+++ b/CCP.Avalonia/App.axaml.cs
@@ -28,8 +28,12 @@ namespace ConditioningControlPanel.Avalonia
         public override void OnFrameworkInitializationCompleted()
         {
 
+            // The app shell is the startup window. Until now this head opened the diagnostics
+            // MainWindow, which was right while the shell did not exist and is wrong now that it
+            // does. The diagnostics window is still reachable, from Settings, and RenderProof
+            // still hosts single views inside it - neither depends on it being the startup window.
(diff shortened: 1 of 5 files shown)


CodeBambi (commented):
Avalonia-only, nothing on the WPF side to review. Under the agreed workflow this belongs on `avalonia/main` (see #637), not in a chain that ends at this repo's main. Retarget when the branch exists.

Issue #637, opened 2026-09-03
Port coordination (WPF x Avalonia)


CodeBambi:
Shared log between the WPF side (CodeBambi) and the port side (@ObviouslyNotMich). Rules live in `ConditioningControlPanel/CLAUDE.md` under "Cross-Platform Port Collaboration" (PR #636).

**When you start a session, post one comment with three lines:**

```
merged: <what landed since your last note>
in flight: <branches / PRs you are working on>
touching next: <files under the OTHER side's directories you will need, or "none">
```

**Standing rules (short version)**
- Port work goes to `avalonia/main`, merged by Mich without review. One merge-up PR `avalonia/main -> main` per week, additive files only, reviewed as a unit.
- Seam PRs (both sides touched): under 200 lines, based on `main`, label `seam`, reviewed within a day. #618 is the first one and should be retargeted to `main`.
- `/Assets` move (#634) waits until the day after 6.9.2 is tagged, and its Windows CI (RG1000 duplicate `resources/comehere.gif`) must be green first.
- 600-line cap per PR on both sides. The weekly merge-up is the only exception.
- Milestone 1: Avalonia shell launches on Linux, logs in, runs a flash session. Proof = green CI + a short clip.

**WPF side, 2026-09-03**
merged: v6.9.1 (73e4b6a), moderation age filter #477
in flight: 6.9.2 hotfix wave, branches `fix/692-*` (VideoService, WebcamTrackingService, MainWindow.xaml.cs theme refresh, MediaCommand/AiCommandService + App.xaml.cs scaffold lines, MainWindow.LabTab.cs + PlayTabView.xaml, FlashFeatureControl.xaml + 9 language files, Resources/web/arcademy/shell)
touching next: nothing under CCP.Core / CCP.Avalonia / CCP.VR

CodeBambi:
Reviewed the whole wire chain today (#618, #634, #635, #638, #649, #650, #660, #661, #662, #663). Per-PR notes are on each PR. This is the process part, and the plan I want us to follow.

**Where we are**

- All ten are stacked on the 139 draft port/* PRs, so "mergeable" on each one means clean against its parent, not against main. Nothing in the chain can land on main by itself right now.
- There is no `avalonia/main` branch yet, no PR carried the `seam` label (I added it to the WPF-touching ones today), and #618 still targets the port chain instead of main.
[...]

PR #638, opened 2026-09-03
feat(avalonia): shell navigation - ShowTab, the door accordion, every rail handler

ObviouslyNotMich:
The shell drew but nothing responded: all 86 handlers the XAML names were empty. This restores the part of `MainWindow.TabNavigation.cs` that is navigation, so every door and rail entry switches tabs.

## Real
- **`ShowTab`**: hide every tab panel, show one. Case-insensitive at the door, for the reason WPF is: a deep link or a mod saying `Settings` must not land on a blank page. An unknown key keeps the current tab.
- **The door accordion**: the door that owns the tab unfolds, the others fold. WPF's `NavDoorMap` copied row for row.
- **Every rail and door click handler**, each the one `ShowTab` call it is in WPF.
- Panels and doors resolve by `x:Name`, so a panel this head lacks is skipped rather than a compile error.

## Not yet, each named in the file header
Transition choreography and the door height tween (panels snap open and shut), per-tab side effects that reach `App.*` or a service, the bark and EmiDesk hooks, the three keys that are windows rather than tabs (`patreon`, `fyp`, `justdrop`), and an active state on the rail, for which no selector exists on this head yet.

## The check
`--nav-check` constructs the shell headlessly, clicks through `ShowTab`, and asserts exactly one tab is visible and the right door is open, including the alias and unknown-key cases. The render proofs are still frames; this is the one proof that exercises a click.

## Verified
Solution builds, Core guards pass, `--smoke` and `--nav-check` pass, 186/186 views render.

### DIFF
--- a/CCP.Avalonia/NavCheck.cs
+++ b/CCP.Avalonia/NavCheck.cs
@@ -1,0 +1,48 @@
+using System;
+using Avalonia.Controls;
+using ConditioningControlPanel.Avalonia.Views.Windows;
+
+namespace ConditioningControlPanel.Avalonia
+{
+    /// <summary>
+    /// The smallest thing that fails if shell navigation breaks. Constructs the shell headlessly,
+    /// calls ShowTab the way every rail handler does, and asserts what the user would see.
+    /// </summary>
+    internal static class NavCheck
+    {
+        public static int Run()
+        {
+            RenderProof.EnsureSetUp();
+            var w = new MainShellWindow();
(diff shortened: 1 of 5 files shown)


CodeBambi (commented):
Avalonia-only. The ShowTab and door-map port from the WPF window reads correctly, the `--nav-check` mode is a nice smoke, and the four window keys that no-op are the right ones for now. Same as #635: retarget to `avalonia/main` once it exists (#637).

PR #642, opened 2026-09-03
fix(video): don't start a video on top of the previous clip's wedged Stop()

CodeBambi:
**Stacks on:** `main` (this is PR 1 of a 3-PR stack in `VideoService.cs`; PR 2 bases on this branch).

## Bug
`CC-Labs-llc/ccp-bugs#1121` - fullscreen goes black for 2-3 seconds and then the video is gone, on some rigs for every video. Reporter is single-monitor, 6.8.4 through 6.9.1, survives a reinstall.

## Root cause
A LibVLC `Stop()` can stay inside native code for seconds on a rig whose audio output resolves no device (the wedge cluster #765/#766/#767 already documents 4936ms/4978ms measurements in this file).

- `VideoService.cs:7043-7085` (`CloseAll`): the vmem/blur path - which is the DEFAULT render path - deliberately skips the 500ms + 4s `Stop()` waits and hands the tasks to an async quarantine. The `VideoView` path gives up after ~4.5s. Either way nothing prevented the NEXT video from being built while those tasks were still running.
- `VideoService.cs:3037` (`CreateLibVLCVideoWindow`): the new `MediaPlayer` is constructed from the SAME shared instance the wedged `Stop()` is still inside. `Play()` returns "playing" and no frame is ever presented.
- `VideoSurfaceHealth.cs:98` `AllowsFrameRetry(primary, armed) => !primary || armed > 1`: the blur-frame watchdog (`VideoService.cs:3690-3700`) therefore gave a LONE primary no re-Play rung at all, so the one surface that could have recovered on a single-monitor rig was the only surface denied a retry.

## Fix
- `CloseAll` publishes its outstanding stop tasks plus the owning instance (`_pendingStopTasks` / `_pendingStopOwner`).
- New `AwaitPendingStops()`, called from `StartVideoPlayback` before `EnsureLibVLCInitialized()`: waits up to 2s, pumping (same `WaitWithMessagePump` the teardown already uses, so the dispatcher keeps draining). If the straggler is still running it does not fight it - `RetireSharedLibVLC(..., fromCurrentPlayback: false)` quarantines the instance and a fresh one is built for this clip. The instance is deliberately NOT disposed: this file's `QuarantineNative` contract is that disposing an object with a wedged native call inside it is the #559 poisoning, so "abandon on a background thread" is implemented as the existing quarantine plus the existing background rebuild.
- One `Warning` line per overrun carrying the elapsed ms and the outcome, plus a matching video-diag breadcrumb.
- `AllowsFrameRetry` now grants a lone primary the same single rung (`armedSurfaces > 0`); zero armed surfaces still gets none (a watch judged before it was registered has nothing to re-Play).

Circuit-breaker cap left at 4 as instructed.

## Verified
- `dotnet build -c Debug` in `ConditioningControlPanel/`: **0 errors**, 344 warnings (pre-existing CA1416/CS8602 noise).
- Test run for the whole stack is reported on PR 3 (single run at the end of the stack, as agreed).
- NOT verified: no repro rig. The overlap window is native timing that cannot be reproduced here; the `CCP_FAULT_WEDGE_STOP=1` DEBUG fault injector in this file is the intended way to exercise it, and that needs a real app launch (not allowed for this agent - the app is single-instance and the owner has one running).
- Trade-off to be aware of: on a rig whose decoder really is dead, the lone-primary rung moves the skip from ~8s of black to ~16s. It buys back every run where the decoder was merely late, which is the common case now that a previous clip's `Stop()` can still be holding the instance.

## Size
102 insertions, 24 deletions across 3 files (`git diff --stat origin/main...HEAD` -> `3 files changed, 102 insertions(+), 24 deletions(-)`).

https://claude.ai/code/session_01XF7ZmKyHYMRoErLBhS2x49

### DIFF
--- a/ConditioningControlPanel/Services/Video/VideoService.cs
+++ b/ConditioningControlPanel/Services/Video/VideoService.cs
@@ -165,6 +165,17 @@ namespace ConditioningControlPanel.Services
         // it treats the player as wedged and quarantines it. Only a real wedge should ever spend a
         // retire from the per-session budget.
         private const int StopStragglerGraceMs = 4000;
+        // Stop() tasks from the LAST teardown that were still running when CloseAll returned, plus
+        // the instance they belong to. The vmem/blur path deliberately never waits for them on the
+        // dispatcher and the VideoView path gives up after ~4.5s, so the NEXT video can start while
+        // a previous Stop() is still inside native code. A Play() that overlaps a wedged Stop() on
+        // the same instance never presents a frame: "fullscreen goes black for 2-3 seconds and then
+        // the video counts as watched" (#1121).
+        private static volatile Task[]? _pendingStopTasks;
+        private static LibVLC? _pendingStopOwner;
+        // How long the next video waits (pumping) for those stragglers before it abandons the owning
+        // instance and builds its players on a fresh one.
+        private const int PendingStopWaitMs = 2000;
 
 #if DEBUG
         // Fault injection for the wedge cluster (#765/#766/#767). Set CCP_FAULT_WEDGE_STOP=1 and the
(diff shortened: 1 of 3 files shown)


CodeBambi:
Doubt 2 settled from the #1121 report log: the clip went through LibVLC (SURFACE: engine=libvlc, role=primary, firstFrame=125ms), then 'MediaPlayer quarantined: Stop() never completed - 60000ms cooldown' twice and 'RETIRE: shared LibVLC retired 4/4'. The browser engine did not take the clip on this rig, so this PR is on the reporter's path.

PR #649, opened 2026-09-03
refactor(core): move the art-framing table into Core on a Core viewbox type

ObviouslyNotMich:
`ModArtFraming` typed its shipped viewboxes as a bare `Rect`, which resolves to `System.Windows.Rect`: the trap `CLAUDE.md` names. That one type was what kept the framing table, the mod manifest that carries it, and behind them the 8,164-line settings model in the WPF head. This is the first of the layers that unblock the settings move (#618 explains the chain).

## What
- **`ArtViewbox`**, a readonly record struct in Core with the same four members. Deliberately not a geometry type: no operators, no intersection, no union. Nothing needs them.
- **`ModArtFraming.cs` moves to Core** with the type swapped. The table and the resolve rules are pure data and pure arithmetic; the file is otherwise unchanged and its namespace is the same, so no consumer changed a `using`.
- **Every consumer but three only reads `Width` and `Height`**, which the record shares, so they compile untouched. The three that hand a value to a WPF `ImageBrush` convert at that boundary with `ToRect()`, one five-line extension in the head.
- The registry's tests follow: seven literal rects and two helper signatures.

## Verified
Core builds and the guards pass (no `System.Windows` reached Core), the WPF head and the Windows test project build, `--smoke` on both heads, `--nav-check`, 186/186 views render. 87 changed lines plus one rename.

### DIFF
--- a/CCP.Core/Models/ArtViewbox.cs
+++ b/CCP.Core/Models/ArtViewbox.cs
@@ -1,0 +1,18 @@
+namespace ConditioningControlPanel.Models
+{
+    /// <summary>
+    /// A normalized viewbox: the sub-region of a source image an art surface shows, in 0..1
+    /// units of that image. X and Y are the top-left corner, Width and Height the extent.
+    ///
+    /// <para><b>Why this and not <c>System.Windows.Rect</c>.</b> The framing table and the
+    /// resolve rules are pure data and pure arithmetic, and they belong in Core next to the mod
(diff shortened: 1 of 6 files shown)


CodeBambi (commented):
Clean. `ArtViewbox` as a readonly record struct in Core plus the WPF `ToRect()` extension is the right split. One leftover `using System.Windows;` in the updated test, not blocking.

Please rebase onto main as a standalone `seam` PR (it is 120 lines) and I will merge it.

PR #650, opened 2026-09-03
feat(core): the settings seams - secrets, change hook, mode name, subreddit rule

ObviouslyNotMich:
The 8,164-line settings model reached the WPF head in four ways. Each is now a Core seam, so the model itself can move in the next layer as a pure `git mv`.

| Was | Now |
|---|---|
| `App.Bark?.NotifySettingChanged(name)` | `CoreSettingsHooks.NotifySettingChanged(name)` |
| `App.Mods?.GetModeDisplayName()` | `CoreMods.ModeDisplayName` |
| `SecureApiKeyStore` / `SecureAuthTokenStore` | `CoreSecrets.Retrieve` / `Store`, by name |
| `FypOnlineCoordinator.SanitizeSub` | `Services.Fyp.SubredditName.Sanitize` |

## The one that needs a reviewer's eye: secrets
The two stores stay in the head. They are DPAPI, which the Core guards forbid because it throws off Windows. **Unseeded `CoreSecrets` means "no store"**: `Retrieve` returns null and `Store` is a no-op. It never means "store in the clear". The Linux head is in that state today, so on Linux the API key and account token are not persisted at all, and the seam's doc says so. A keyring-backed provider is a later layer; plaintext on disk is not a fallback.

## The rest
- The subreddit rule is pure string work that lived in the online coordinator by accident. It moves to Core; the coordinator delegates, so its callers are unchanged.
- It had no test of its own, so the Linux smoke runner now checks its six edges, plus the unseeded behaviour of every seam above.
- Comments in the model that named `App.*` are reworded, so the guard that forbids `App.*` in Core has nothing to catch when the file moves.
- The WPF head seeds all four in `App`, next to the existing `CoreMods` seeding. Bark and the mod service attach later in startup; the delegates read them lazily, so the order is safe.

## Verified
Core builds and guards pass, the WPF head and Windows tests build, the Linux Core smoke runner passes with the new checks, `--smoke` and `--nav-check` on the Avalonia head, 186/186 views render. About 180 changed lines.

### DIFF
--- a/CCP.Core/CoreMods.cs
+++ b/CCP.Core/CoreMods.cs
@@ -56,6 +56,16 @@ namespace ConditioningControlPanel
             get { try { return PetNameOverrideProvider?.Invoke(); } catch { return null; } }
         }
 
+        /// <summary>The active mod's display name for its content mode, or null when no mod layer
+        /// is up. The settings model shows it and falls back to the vanilla label.</summary>
+        public static volatile Func<string?>? ModeDisplayNameProvider;
+
+        /// <summary>Mode display name, or null. Faults are swallowed - see <see cref="ActiveModToken"/>.</summary>
+        public static string? ModeDisplayName
+        {
+            get { try { return ModeDisplayNameProvider?.Invoke(); } catch { return null; } }
+        }
+
         /// <summary>Collective override, or null. Faults are swallowed - see <see cref="ActiveModToken"/>.</summary>
         public static string? CollectiveOverride
         {
--- a/CCP.Core/CoreSecrets.cs
+++ b/CCP.Core/CoreSecrets.cs
@@ -1,0 +1,43 @@
+using System;
+
+namespace ConditioningControlPanel
+{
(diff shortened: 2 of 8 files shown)


CodeBambi (changes requested):
Seams are fine and the subreddit sanitiser move is good. One real fix before merge:

**Legacy OpenRouter key is lost on a head with no secret store.** The `AppSettings.OpenRouterApiKeyLegacy` setter now does `CoreSecrets.Store`, which is a no-op when nothing is seeded. The getter still returns null, so the next save strips the plaintext key from settings.json and the user's key is gone. #661 fixes exactly this for the auth token with a `CoreSecrets.HasStore` guard, but that guard never got applied here. Move `HasStore` into this PR and skip the migration when it is false (leave the plaintext in place until a store exists).

Nit: the rewritten comment in `OnPropertyChanged` reads "...props. the bark service is null during startup".

Then rebase onto main as a `seam` PR and I will merge it.

PR #660, opened 2026-09-03
refactor(core): move the settings model and its closure into Core

ObviouslyNotMich:
`AppSettings` (8,164 lines) and the eleven files it cannot compile without move into Core: `RampMath`, `Preset`, `HapticSettings`, `CornerGifOverlaySetting`, `AwarenessIntensity`, `AssetPreset`, `PhrasePreset`, `ModManifest`, `Session`, `SessionDefinition` and `BuiltInMods`. Pure `git mv`, namespaces unchanged, no consumer edited a `using`. 67 changed lines beside the renames.

## Five reaches surfaced on the way, each now a seam
| File | Was | Now |
|---|---|---|
| `PhrasePreset`, `AssetPreset` | `App.Settings.Current` | `CoreSettings.Current` |
| `Session` | `App.Mods?.MakeModAware(x) ?? x` | `CoreMods.MakeModAware(x)` |
| `Session` (spoiler text) | `App.Settings?.Current.LockCardPhrases` | `CoreSettings.Current.LockCardPhrases` |
| `AwarenessIntensity` | `App.Settings?.Current?.AwarenessIntensity ?? Chatty` | `CoreSettings.HasProvider ? … : Chatty` |

**`CoreSettings.Current` is never null.** With no head attached it is one shared default instance: the honest "nothing loaded" state under tests, the smoke runner and early startup. `HasProvider` says which case you are in, and `AwarenessIntensity` uses it to keep its documented "unseeded reads as the ship default, not as Off" rule exactly. The WPF head seeds both seams in `App`.

## What deliberately stays in the head
`HapticMixer` drives real devices and needs `HapticDeviceManager`, which logs through `App`. `HapticSettings` only wanted the mixer's three default constants; those now live with the settings (`HapticSettingsV2`) and the mixer aliases them, so its public surface is unchanged.

## Verified
Core builds and the guards pass, the WPF head and Windows tests build, the Linux Core smoke runner checks the unseeded behaviour of every new seam, `--smoke` and `--nav-check`, 186/186 views render.


CodeBambi (commented):
The code is equivalent (pure renames, `CoreSettings.Current` never null, `MakeModAware` moved). One nit and one hold.

**Nit:** in `HapticSettingsV2` the new summary for `DefaultMasterCap` / `DefaultSoftRampMs` / `DefaultMaxConcurrentPulses` was inserted between the existing "SAFETY" summary on MasterCap and the property, so MasterCap now carries two summary blocks and the constants read as its docs. Move the constants above the existing comment.

**Hold:** this moves AppSettings.cs (134 commits in the last 60 days, 32 in the last 14) into CCP.Core, which the agreement puts under your ownership, and two open 6.9.2 branches already edit it. The ownership call is in #637 (shared folder, WPF side edits directly, you only via seam PRs, recipe in ConditioningControlPanel/CLAUDE.md updated in this PR). This lands after 6.9.2 is tagged, behind #634.

PR #661, opened 2026-09-03
refactor(core): move SettingsService into Core; the Linux head has real settings

ObviouslyNotMich:
`SettingsService` moves into Core as-is, with every reach into the WPF head through a seam. From this layer the Linux app reads and writes `settings.json` under `~/.local/share/ConditioningControlPanel`, with the same migrations, daily backups and interrupted-save recovery the Windows app has.

| Was | Now |
|---|---|
| `App.Logger?.X(…)` (45 sites) | Serilog's static `Log.X(…)`. Same sink: `App` assigns `Log.Logger`. |
| `App.UserDataPath` | `CorePaths.UserData` |
| `SecureAuthTokenStore` | `CoreSecrets`, by name |
| `App.HasCloudIdentity && App.ProfileSync != null` → `BackupSettingsAsync` | `CoreSettingsHooks.CloudBackup`, null on a head with no cloud. The 401-storm throttle stays in the service. |
| `KeywordTriggerService.RebuildActionsFromFlatFields(t)` | `t.RebuildActionsFromFlatFields()`: pure model logic, now on `KeywordTrigger` in Core; the service's static delegates to it. |

## Two behaviour fixes, both found by making it run off Windows
- **The plaintext-to-encrypted token migration runs only when a secret store is attached.** Without one it would strip the plaintext on re-save and lose the token. On the Linux head today it leaves the value where it was, which is what `CoreSecrets` promises.
- **The service creates its own folder before writing.** The WPF head happens to create `UserData` at startup. On a fresh profile anywhere else, the first save threw `DirectoryNotFoundException` into a sink nobody was reading, and the smoke check below is what caught it.

## Both heads
The WPF head seeds `CloudBackup` next to the other seams. The Avalonia head constructs the real service under the desktop lifetime and seeds `CoreSettings` from it, deliberately **not** in `Initialize()`: the headless render path never reaches that callback, so a CI render cannot touch a user's profile, and the renders keep binding against the default instance.

## The check
The Linux smoke runner now sandboxes `CorePaths` through `CCP_USERDATA_DIR`, writes a value through the real service, and reads it back from a fresh instance. It failed on the first run, which is how the folder bug surfaced.

## Verified
Core builds and guards pass, WPF head and Windows tests build, Linux Core smoke passes with the round-trip, `--smoke` and `--nav-check`, 186/186 views render. 245 changed lines plus one rename.

### DIFF
--- a/CCP.Avalonia/App.axaml.cs
+++ b/CCP.Avalonia/App.axaml.cs
@@ -2,11 +2,15 @@ using Avalonia;
 using Avalonia.Controls.ApplicationLifetimes;
 using Avalonia.Markup.Xaml;
 using ConditioningControlPanel.Localization;
+using ConditioningControlPanel.Services;
 
 namespace ConditioningControlPanel.Avalonia
 {
     public partial class App : Application
     {
+        /// <summary>The settings service, or null on the headless render path.</summary>
+        public static SettingsService? Settings { get; private set; }
+
         public override void Initialize()
         {
             AvaloniaXamlLoader.Load(this);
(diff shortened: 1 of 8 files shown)


CodeBambi (commented):
Good PR. The Serilog static `Log` swap is safe (App assigns `Log.Logger` at startup), the HasStore guard on the auth-token migration is exactly right, and `Directory.CreateDirectory` before every save is a real fix. The Linux smoke round-trip is welcome.

**One change:** the cloud-backup gate moved inside the delegate. WPF seeds `CoreSettingsHooks.CloudBackup` unconditionally and returns a completed task when there is no cloud identity, but the service bumps `_lastBackupAttemptTicks` before calling it. So the throttle window is consumed on every save with nobody logged in, and the first real backup after login can be delayed by a full window. The old code checked identity before the throttle. Seed the delegate as null when there is no identity, or add a "has identity" hook the throttle checks first.

Nit: stray blank lines around `KeywordTrigger.RebuildActionsFromFlatFields`.

Same hold as #660 on the SettingsService.cs move (ownership, #637). Lands after 6.9.2.

PR #663, opened 2026-09-03
feat(core): the mod seams - what views ask of the mod service, and what it asks the head

ObviouslyNotMich:
The mod service stays in the WPF head. I tried the move: on a switch it clears caches on a dozen head services (`ModCompanionContent`, `ModResourceResolver`, `SessionEngine`, `BambiSprite`, two web hosts, the updater, the phrase service…), and moving that is a project, not a layer. Two seams go around it instead.

## What views ask of it
`CoreMods` gains `ActiveModId`, `IsCCPDefault`, `InstalledMods`, `AccentColorHex`, `SecondaryColorHex`, `Affirmation`, `GetPhrases`, `PinkRushName` and the `ModChanged` event: every member the ported views' stubs name. **Unseeded, each answers exactly what the service answers with no mod active**: the built-in CCP default manifest, which is already in Core. A head with no mod layer (Linux today) draws vanilla rather than nothing, and the smoke runner checks every answer against that manifest rather than against a literal.

## What it asks the head
`CoreModsHooks` (brain reset, bark reload, live-event accent, active companion, switch companion) and `CoreReleaseContent` (pack ids, install stamp, pack info, the pack-installed event) replace the direct `App.*` reaches the service made on a switch. The head seeds all of them next to the other seams and forwards `ModChanged` into Core. `AttachReleaseContent` no longer takes the pack service; it subscribes to the Core event.

## Moves
`ContentLocator` (logger swap only) and `ModPackage` move to Core as-is.

## Verified
Core builds, guards pass, both heads and the Windows tests build, Linux Core smoke with the new checks, `--smoke`, `--nav-check`, 186/186 views render. 221 changed lines plus two renames.


CodeBambi (changes requested):
The seam surface is right and the fallback colours match ModService on main. ModService is a single instance so the static `PackInstalled` subscription that never unsubscribes is fine today. Three fixes before merge:

**1. Dead guard in `ModService.Initialize`:**

```csharp
if (CoreSettings.HasProvider) App.Settings.CurrentReplaced += OnSettingsReplaced;
```

The App static ctor seeds a non-null delegate whether or not `App.Settings` exists yet, so `HasProvider` is always true here and the old null check is gone. Use `CoreSettings.Service is { } svc` and subscribe on that. The `CoreSettings.Current != null` checks elsewhere in this PR are dead too since Current is never null.

**2. One try block covers both events in `ReleaseContentService`:**

```csharp
try { PackInstalled?.Invoke(this, packId); CoreReleaseContent.RaisePackInstalled(packId); } catch { }
```

If a WPF subscriber throws, the Core event never fires and the Avalonia side never hears about the pack. Guard the two calls separately, and please expand the braces.

**3. Leftovers:** `// == ContentLocator.ContentRoot` on ContentLocator.ContentRoot itself, and `_log = Log.Logger` captured at type init pins whatever logger existed at first touch.

Lands after 6.9.2 with the other moves (#637).

PR #793, opened 2026-09-04
feat(arcademy): the slip, part 4 - the delivery ladder

CodeBambi:
Stacked on #792 -> #791 -> #790. **Review in that order; do not merge ahead of the stack.**

## What

#792 gave the card a button and a download. This gives it the rungs that should be tried *before* the floor.

| # | Rung | Who it is for | Why it is gated the way it is |
|---|------|---------------|-------------------------------|
| 1 | `navigator.share({files})` | phone | Gated on **`canShare({files})`**, not on `share`. Every host that can share text has `navigator.share`; handing one of those a file list gets you a rejected promise instead of a sheet. A closed sheet is an `AbortError` and counts as **done** - falling through to the clipboard after someone deliberately backed out would be rude. |
| 2 | `ClipboardItem` | desktop browser | The **promise form** on purpose: Safari only honours a clipboard write issued inside the gesture, and takes a pending promise precisely so a slow encode does not lose it. Chrome accepts the same shape, so there is one code path. |
| 3 | the host bridge | desktop app | WebView2 gives the page no async clipboard image write worth the name, so the PNG goes over the bridge as base64 and C# puts it on the Windows clipboard itself. |
| 4 | `<a download>` | everyone | The universal floor - and the rung the Discord Activity iframe can **silently refuse**: its sandbox drops navigations it did not start, and there is no event for that. |
| 5 | nothing, **said out loud** | the Activity, mostly | "Sharing is not available here". A share that quietly did not happen is the worst of the five outcomes, not the least. |

Every rung is feature-tested before it is tried and every rung is wrapped, so a host that *has* an API and refuses anyway falls to the next one rather than taking the button down with it.

The busy latch now has **one owner on every path** - the ladder is two awaits deep, and a beat left running because a rung threw is a button nobody can press again tonight.

## Host side

`share-image` decodes the base64, checks the PNG signature before anything is asked to decode it, and does both the decode and the `Clipboard.SetImage` **on the UI thread** - WPF's clipboard is STA, and an unfrozen `BitmapImage` handed across threads is a crash waiting for a slow night. One retry on `CLIPBRD_E_CANT_OPEN`, which is what every other app on the machine does. Payload capped at ~4.4 MB of text (the page caps itself well under that; this is the wall behind it). **Exactly one reply always goes back**, `ok` true or false, because a missing reply is a spinning button.

`shareToHost` is handed to the report card only on the app - on the web that rung is *absent*, not present-and-broken.

## Test evidence

- Node suite: **104 passed, 0 failed** (12 new, covering the ladder's ordering, the throw-through, the oversized-payload refusal, the missing-bridge case, and that the module carries no strings of its own).
- `node --check` clean on all five touched JS files.
- `dotnet build`: **0 errors**.

**Exercised for real:** the ladder's ordering and every refusal path, in node against stubs; the C# compiles.
**Not exercised:** an actual `navigator.share` sheet on a phone, an actual `ClipboardItem` write in Safari or Chrome, an actual round trip to `Clipboard.SetImage` through WebView2, and the Activity iframe's behaviour. All four need a device or the running app in hand. Every one of them is written to fail *down the ladder* rather than sideways, and rung 5 exists so the last failure is spoken rather than silent.

https://claude.ai/code/session_016cbafJm7hrM8qf4Fr7Wj6r

### DIFF
--- a/ConditioningControlPanel/Resources/web/arcademy/boot.js
+++ b/ConditioningControlPanel/Resources/web/arcademy/boot.js
@@ -867,6 +867,9 @@ bridge.on('profile', guard('profile', (m) => {
 }));
 bridge.on('meta', guard('meta', (m) => { if (shell) shell.onMeta(m); }));
 bridge.on('annex-stats', guard('annex-stats', (m) => { if (shell && shell.onAnnexStats) shell.onAnnexStats(m); }));
+bridge.on('share-image-result', guard('share-image-result', (m) => {
+  if (shell && shell.onShareImageResult) shell.onShareImageResult(m);
+}));
 bridge.on('fullscreen', guard('fullscreen', (m) => { fullscreen = !!m.on; }));
 bridge.on('ping', guard('ping', (m) => bridge.send({ type: 'pong', t: m && m.t })));
 bridge.on('end-run', guard('end-run', () => shutdown('host asked')));
--- a/ConditioningControlPanel/Resources/web/arcademy/core/lexicon.js
+++ b/ConditioningControlPanel/Resources/web/arcademy/core/lexicon.js
@@ -87,6 +87,8 @@ export const DEFAULT_LEXICON = Object.freeze({
   share_image: 'Share report card',
   share_add_name: 'Add my name',
   share_saved: 'Report card saved',
+  share_copied: 'Report card copied',
+  share_shared: 'Report card shared',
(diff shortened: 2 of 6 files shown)


CodeBambi:
Superseded by #799 (base branch was deleted, GitHub would not retarget).

PR #797, opened 2026-09-04
feat(arcademy): Back Room D1 - chips shape, casino bridge, cage adoption

CodeBambi:
The desktop half of "The Back Room" wave 1: the chips currency exists in the wallet shape, the host can talk to the casino routes, and a cage press moves Sparkle Points for real. No UI, no cabinets, no catalog rows here (PR D2 owns the shelf, W1 the door, K1 the kit).

Built to `BACKROOM-CONTRACT.md` v1, sections 1, 3 and the C# side of 6.

## What and why

**`ArcademyEconomy`** gains `CurChips = "c"` and `ChipsPerSparkle = 100`. `EnsureShape` now ensures integer `c` and `earnedC` beside `t`/`earnedT`, and `BalanceJson` emits both. Chips ride the existing wallet blob so one adopt carries all three purses, and an older build that has never heard of chips still round-trips them rather than zeroing what the server holds. No catalog rows and `CurrentWave` is untouched.

**`ArcademyWalletSyncService`** gains one generic call:

```csharp
public static void Casino(string op, JObject body, string localDay, Action<CasinoOutcome> settled)
```

- `status` -> `GET /v2/arcademy/casino?unified_id=&localDay=`
- `cage` -> `POST /v2/arcademy/casino/cage`
- `play` -> `POST /v2/arcademy/casino/play` (body gains `localDay`)

Same `X-Auth-Token` + `unified_id` door, same 5s `PageWait`, same generation guard and the same tier-refusal mapping `PostBuyAsync` uses. `CasinoOutcome(bool Answered, int Status, JObject? Body)`; a transport failure, the identity door and a dead token are all `Answered false / Status 0`. Nothing is adopted in that lane: a casino answer carries balances, not a wallet, so it stays a pure courier and the host decides what landed. The page's `localDay` is shape-checked against `yyyy-MM-dd` and falls back to this desk's own date, because the free scratcher is bounded by it.

**`ArcademyHostService`** handles the two new page messages. **`ArcademyMetaStore`** gains `NoteChips(int? chips, int? earnedChips)`, a targeted write: adopting a balances-only answer as a whole wallet would blank the tickets, the shelf and the two lever rungs.

**`ProfileSyncService`** stops max-merging skill points on the v2 path. Explained below.

## Bridge message shapes

Up, from the page:

- `{ type: 'casino-request', reqId, op: 'status'|'cage'|'play', body, localDay }`
- `{ type: 'triggers-request' }`

Down, from the host, exactly one per request:

- `{ type: 'casino-result', reqId, ok, status, body }`
  - offline / unknown op / no account: `ok:false, status:0, body:{ reason:'offline' }`
  - tier refusal: `ok:false, status:403`, and the server's own 403 body passes through verbatim
  - a game refusal is the server's 200 with `ok:false` and rides back whole
  - `ok` is true only for a 2xx whose body says `ok:true`
- `{ type: 'triggers-result', triggers: [...] }` from `AppSettings.CustomTriggers`, raw text, empties dropped, max 64 entries each trimmed to 64 chars

On `op:'cage'` with `ok`, the host adopts `body.sparkle` into `App.Settings.Current.SkillPoints` and calls `TrackSkillPointsSpent(body.credited / 100)` (the cage answers in chips credited, the prestige counter counts Sparkle spent). On `cage` and `play` alike it writes the answer's `chips` / `earnedC` into the wallet cache and posts the meta snapshot, so the balance chip, the cage sign and the Prize Counter repaint from one frame.

## The ProfileSync change

`/v2/user/sync` used `Math.Max(server, local)` for skill points, with a comment saying the balance only ever increases. That was true while nothing could spend it outside the server's own purchase route being immediately adopted. It is not true now: the cage DEBITS `skill_points` server side, and a max-merge would resurrect every debit the moment a second desk synced its older, higher number. That is chips bought for free, and the skill tree has the same hole today.

So the v2 path now ADOPTS what the server sent. This is safe for the offline level-up case that motivated max in the first place: the server computes the level award and the bubble award FROM the deltas carried up by that same request, so the number being adopted already contains them. `force_skills_reset` keeps its own branch untouched. The legacy path keeps `Math.Max` and now carries a comment saying why the two differ (that endpoint knows about no debit at all).

## Build

`dotnet build` from `ConditioningControlPanel/`: **0 errors**, 322 warnings, all pre-existing (CA1416 platform, CS8602 nullable, WFAC010 manifest). No warning from any file this PR touches. App not run.

Changed lines: 378 added / 15 removed across 5 files, inside the ~600 cap.

## Merge note

**Needs server S1 on the wire first.** Until `/v2/arcademy/casino*` exists, every `casino-request` answers `{ok:false, status:404}` or the offline shape, which is exactly what the page is built to survive but is not something to ship as a working room. Contract merge order: cclabs-web parser and host relay, then server S1, then this. The `ProfileSyncService` adoption is independently correct and can land at any time.

### DIFF
--- a/ConditioningControlPanel/Services/Arcademy/ArcademyEconomy.cs
+++ b/ConditioningControlPanel/Services/Arcademy/ArcademyEconomy.cs
@@ -24,7 +24,7 @@ namespace ConditioningControlPanel.Services.Arcademy;
 /// for the same reason the streak does: a stale or edited page must never be able to mint currency.
 /// Shape (all fields ensured by <see cref="EnsureShape"/>):</para>
 /// <code>
-/// { "t":0, "k":0, "earnedT":0, "earnedK":0,
+/// { "t":0, "k":0, "c":0, "earnedT":0, "earnedK":0, "earnedC":0,
 ///   "tokenDays":["yyyy-MM-dd"],
 ///   "payDays":{ "yyyy-MM-dd": { "&lt;gameKey&gt;": 2 } },
 ///   "inv":{ "late_slip": { "n":2, "at":"yyyy-MM-dd" } },
@@ -236,6 +236,16 @@ internal static class ArcademyEconomy
     public const string CurTickets = "t";
     public const string CurTokens = "k";
 
+    /// <summary>CHIPS, the Back Room's third purse. Bought one way at the cage with Sparkle Points
+    /// and spendable only on the Back Room shelf: chips never cash out and nothing else ever mints
+    /// them, so the school's two earned currencies keep their meaning.</summary>
+    public const string CurChips = "c";
+
+    /// <summary>The cage's one and only rate. One number, here, because the page prints it, the
+    /// counter quotes it and the server debits by it, and a rate that disagrees across those three
+    /// is a player watching Sparkle vanish into the wrong pile.</summary>
+    public const int ChipsPerSparkle = 100;
+
     public const string SkuLateSlip = "late_slip";
 
     /// <summary>How many Tardy Slips the desk will hold — and therefore the longest absence one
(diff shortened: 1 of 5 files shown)


CodeBambi:

PR #803, opened 2026-09-04
feat(arcademy): Back Room K1a - the cashier's line and the money

CodeBambi:
**Merge order: K1a -> K1b -> K1c -> K1d -> K1e.** This is the first of five and the only one based on `main`. Nothing here is wired into the shell yet; PR W1 imports `backroom/index.js`'s `openBackRoom(ctx)`, which lands in K1c.

The two pieces every machine on the casino floor will share, and nothing else. No room, no cage, no cabinets.

### `backroom/kit/api.js` (223 lines)
The promise wrapper over the shell's `send`/`listen`.

- Every answer has the same `{ok, status, body}` shape whether the server said yes, said no, or never answered. No caller needs a try/catch around a rejection, and the floor can draw a dark cage instead of throwing.
- **Missing `send` or `listen` is not an error.** It resolves `offline` on the next tick, which is what a rig, a fixture page and a web host without the relay should see.
- 8s timeout resolves `{ok:false, status:0, body:{reason:'timeout'}}`. A timeout is not a refusal, which is why every id is minted once and reused on a retry.
- Ids are 32 hex characters, inside `ArcademyWalletSyncService.IdShape` (`^[A-Za-z0-9_-]{8,64}$`), because a `cageId` is an idempotency key. `crypto.randomUUID`, then `getRandomValues`, then a clock+counter floor so a hardened webview still mints something legal.
- `localDay()` is `shell/bugle.js`'s six lines copied rather than imported (the kit does not reach into the shell).

### `backroom/kit/chips.js` (261 lines)
The balance plate, the drawn stack and THE BANK, which goes both ways on purpose: a stake leaving is the same ceremony as a win arriving, played backwards. A room where money only ever arrives with a flourish is lying about what it is.

All of it is rAF driven, so the global reduced-motion freeze at the bottom of `styles.css` cannot reach it and every path checks the flag itself and lands on the final number instantly (trap 92).

**Two guards the headless probe earned**, both live bugs it caught:

1. `countUp` reads **one clock at both ends**. The stamp rAF hands its callback is not always on `performance.now`'s origin, and mixing them made `t` land outside 0..1, which painted a **minus balance** for a frame. A balance may never read wrong, not even for one frame.
2. Progress has a **frame-count floor** as well as a clock. A clock that never moves left the loop spinning for ever, which also starved the `setTimeout` belt underneath it: a renderer with a frame permanently pending never goes idle.

### Ledger truth
No chip arithmetic anywhere in either file. The server resolves; these carry the question and the receipt.

### Verification
Both files are exercised by the fixture that lands in K1d (22 checks, all passing on the tip of the stack).

### DIFF
--- a/ConditioningControlPanel/Resources/web/arcademy/backroom/kit/api.js
+++ b/ConditioningControlPanel/Resources/web/arcademy/backroom/kit/api.js
@@ -1,0 +1,223 @@
+/* ============================================================================
+ * backroom/kit/api.js - the cashier's telephone.
+ *
+ * The page never dials the proxy. It says one thing up the seam and waits for
+ * one thing back (BACKROOM-CONTRACT §3):
+ *
+ *   up    { type:'casino-request', reqId, op:'status'|'cage'|'play', body, localDay }
+ *   down  { type:'casino-result',  reqId, ok, status, body }
+ *
+ * Every answer this module hands back has the SAME shape, win or lose or never
+ * arrived: `{ ok, status, body }`. A caller therefore never has to write a
+ * try/catch around a promise that might reject, and the floor can render a dark
+ * cage instead of throwing. That is the whole reason this file exists: a room
+ * with no line to the counter is still a room.
+ *
+ * There is no client-side chip arithmetic here on purpose (LEDGER TRUTH). The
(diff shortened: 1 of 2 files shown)


CodeBambi:

PR #804, opened 2026-09-04
feat(arcademy): Back Room K1b - skin, words and dealer

CodeBambi:
**Merge order: K1a (#803) -> K1b -> K1c -> K1d -> K1e.** Based on `feat/backroom-k1a`.

Everything the room says and everything it looks like, still with no room to hang it on.

### `backroom/backroom.css` (201 lines)
Dark felt under pink and violet neon, scoped entirely to `.bk-room` so it can be lazily linked and cannot leak. Borrows the page's own tokens rather than inventing colours; `--disp`/`--body`/`--mono` only, no serif, no font host; house camera law throughout. Carries the phone diet under `html.arc-mobile` and `html.ae-lite` and the kills under `html.arc-reduced`.

### `backroom/lex.js` (108 lines)
Every string the floor can print, exported as DATA the way `games/impulse-control/lex.js` does it, so `ArcademyHostService.NeutralLexicon` can be mirrored key for key by copying values rather than re-typing them.

- Every row is **under 96 characters**, because `MergeModTable` silently drops a longer one and a dropped row can never be re-voiced by a mod (trap 26).
- The `{0}` slots are filled by `fill()` rather than by template literals, so a translator can move them inside their sentence.
- `makeT(t)` wraps the shell's `t` so a caller cannot forget the English floor.

### `backroom/kit/voice.js` (126 lines)
E.M.I. dealing. 35 lines across seven buckets (`deal`, `win`, `loss`, `big`, `drop`, `royal`, `cage`), and the four rules that separate a dealer from a slot machine with subtitles:

- **Sympathy on losses, never a jab.** The house took the chips; the dealer does not also take a swing.
- **Never promise the next one.** No "you are due". That is the one sentence a room like this must not say, and it is a lie about the odds besides.
- **Diegetic.** A person at a table, not an assistant with a notification.
- **Through the lexicon.** Every line is a key, so a mod re-voices the whole dealer with `lexicon.json`. Nothing here reads `modId`, and nothing may.

### One deliberate copy
`bk_cage_locked` is `prize_tier`'s line verbatim rather than a new sentence. The Prize Counter already says *"The bank does not serve this account yet. Nothing was charged."* when the bank answers 403 with `min_tier`, and the two rooms must not disagree about what that means.

### Separate finding, not fixed here
`prize_tier` itself has **no row in `ArcademyHostService.NeutralLexicon`** (`grep -rn "prize_tier" --include=*.cs` returns zero hits) while all its siblings do. The refusal works, because the JS fallback carries it, but it can never be mod-skinned or localized. That is a live trap-123 hole in the Prize Counter and out of scope for this stack.

### LEXICON KEYS TO MIRROR
The full list of `bk_*` keys and English values for `NeutralLexicon` is in K1e's body, so the integrator has one list covering all five PRs rather than three partial ones.

### DIFF
--- a/ConditioningControlPanel/Resources/web/arcademy/backroom/backroom.css
+++ b/ConditioningControlPanel/Resources/web/arcademy/backroom/backroom.css
@@ -1,0 +1,201 @@
+/* ============================================================================
+ * backroom/backroom.css - the felt, the neon and the brass.
+ *
+ * Lazy-linked by backroom/index.js on first open, exactly like rooms.css. Every
+ * colour is a var() off the shell's tokens or a color-mix of them, so a mod
+ * palette reskins the casino for free and nothing here is a taste nobody can
+ * override. The two literals are the felt greens, and they are literals for the
+ * same reason the card hues are: a casino table is a REMEMBERED colour, not a
+ * derived one, and a violet felt reads as a bug.
+ *
+ * TYPE: --disp for signage, --body for copy, --mono for numbers. No serif
+ * anywhere (house camera law) and no font host (trap 2).
+ *
+ * MOTION: every keyframe in here is decoration and the global freeze at the
+ * bottom of styles.css already flattens it under html.arc-reduced. What this
+ * sheet adds on top is the DIET: html.arc-mobile and html.ae-lite drop the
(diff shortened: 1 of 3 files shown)


CodeBambi:

PR #805, opened 2026-09-04
feat(arcademy): Back Room K1c - the floor and THE CAGE

CodeBambi:
**Merge order: K1a (#803) -> K1b (#804) -> K1c -> K1d -> K1e.** Based on `feat/backroom-k1b`.

**This is the PR that PR W1 needs.** `openBackRoom(ctx)` exists after this one.

### `backroom/index.js` (335 lines) - the floor
A SCENE, like the Annex and the Records Office: no registry row, no timetable deal, no grade, no punch card. Returns `{ root, close, destroy, escapeStep, fit }`, a superset of the `{close()}` the contract asks for, so the shell can hold it the way it holds every other scene.

Paints the header with the chip balance and a lit way out, the pot sign, the cage, and five cabinet frames (`twentyone`, `triple`, `spiral`, `scratcher`, `wheel`).

- Each frame dynamically imports `./<key>.js` at open. Nothing behind it means a **DUST SHEET** that stops being a button, not a console error a player can see. A machine that throws on the way in puts the player back on the floor rather than on a blank stage.
- **Every field of `ctx` is optional**, `send` and `listen` included. A rig, a fixture and a host with no relay all get a room rather than an exception.
- Live stakes come off `status.config.stakes`; the constants in the file are frame placeholders only, so the page never publishes odds of its own.

**Law VI is built first, literally.** The sign exists before the cage does, it is never disabled, and it never waits on a request in flight. `escapeStep()` folds a mounted machine and then answers `false` so the shell walks home. **Nothing here binds a key** - the shell owns the ladder, per `shell/exits.js`'s wiring law. (The fixture in K1d wires Esc the way the shell does and asserts the room detaches.)

### `backroom/cage.js` (224 lines) - THE CAGE
Sparkle in, chips out, 1:100, one way, with the stepper (1 / 5 / 10 / 25 / 50 / custom to 500).

**The tree line is the whole reason this is its own file.** The same sparkle buys skills upstairs, and a casino that quietly competes with the progression tree for the same scarce currency is running a confidence trick. So the counter prints the other price too, every time, in the same breath:

> The same 25 sparkle goes toward soft focus on the tree, 15 short of it.

with `bk_cage_tree_buy` when the amount already covers it and `bk_cage_tree_done` when the tree is finished.

**One press commits.** No confirm dialog, because a modal over a casino counter is a modal over the way out. The guard is a button that stays dead until the number is both sane and affordable. Refusals read warmly and map every reason the contract lists (`insufficient_sparkle`, `invalid_amount`, `busy`, `force_skills_reset`, `arcademy_locked`, offline/timeout/closed) plus any 403 to the standard tier copy. The house never scolds a player for being poor.

### Ledger truth
Neither file adds a chip to anything. The cage hands the answer up and the floor flies it into the plate **from the counter**, so the money is seen to come from the cashier rather than to appear.

### Verification
The fixture lands in K1d. On the tip of the stack, 22 checks pass, including: floor renders with 5 dust sheets, cage sends `{"op":"cage","body":{"sparkle":10,"cageId":"<32 hex>"}}` with `localDay`, the answer banks 1,250 -> 2,250, an unaffordable amount leaves the button dead and sends nothing, the tier refusal prints the standard copy, Esc detaches the room, and **with `send`/`listen` deleted the floor still renders and the cage goes dark without throwing**.

### DIFF
--- a/ConditioningControlPanel/Resources/web/arcademy/backroom/cage.js
+++ b/ConditioningControlPanel/Resources/web/arcademy/backroom/cage.js
@@ -1,0 +1,224 @@
+/* ============================================================================
+ * backroom/cage.js - THE CAGE, and the choice it makes you look at.
+ *
+ * Sparkle in, chips out, one hundred to one, ONE WAY. Nothing on this page can
+ * turn chips back into sparkle, and the cage says so on the wall rather than
+ * letting you discover it at the worst possible moment.
+ *
+ * THE TREE LINE IS THE POINT OF THIS FILE (owner ruling). The same sparkle that
+ * buys chips down here buys skills upstairs, and a casino that quietly competes
+ * with a progression tree for the same scarce currency is running a confidence
+ * trick. So the counter prints the other price too, every time, in the same
+ * breath: "the same 25 sparkle goes toward soft focus on the tree, 15 short of
+ * it". The player then chooses, which is the only version of this worth
+ * shipping.
+ *
+ * ONE PRESS COMMITS. There is no confirm dialog, because a modal over a casino
+ * counter is a modal over the way out (Law VI). The guard is that the button
+ * stays dead until the number is both sane and affordable, and the receipt
+ * afterwards says exactly what moved.
+ *
+ * IT OWNS NO NUMBERS. `paint(snap)` is fed the server's last status frame and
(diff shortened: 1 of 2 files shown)


CodeBambi:

PR #806, opened 2026-09-04
feat(arcademy): Back Room K1d - hold ring, drop beat and the floor fixture

CodeBambi:
**Merge order: K1a (#803) -> K1b (#804) -> K1c (#805) -> K1d -> K1e.** Based on `feat/backroom-k1c`.

Two kit pieces with no cabinet to live in yet, and the fixture that finally lets all of it be driven.

### `backroom/kit/holdring.js` (166 lines)
The one gesture in the room that costs a beat of your attention. A tap is free and a tap is what a habit does; 900ms of holding while a ring closes is a decision.

**Letting go is always free**, on every road out: leaving the button, lifting, blurring, `pointercancel`, releasing the key. Pointer and keyboard both, because a hold that only works with a mouse is a hold half the room cannot use. This is also why the room has no confirm dialog anywhere - **this is the confirm**, and unlike a dialog it never stands between the player and the way out.

**The commit is a `setTimeout` and the ring is only paint.** Hanging the decision off rAF made it hostage to a frame clock, and a tab with no compositor never finished the hold at all. A gesture that silently stops working is worse than one that never existed. Cancelling clears the timer, so the free-release half stays exactly right. Reduced motion keeps the hold and simply does not start the sweep.

### `backroom/kit/dropbeat.js` (161 lines)
The wedge that is not a payout: the room dims, a Loom spiral turns over the felt, the dealer says one thing, the lights come up.

- **The 1.6s ceiling is armed before the first frame is drawn**, so it is enforced rather than intended.
- **The veil is `pointer-events: none`** and takes no focus, so the way out stays clickable underneath it for the whole beat. A ceremony never becomes a modal (Law VI).
- A drop during a drop is **refused, not queued** - two celebrations talking over each other cancel each other.
- Reduced motion draws **one still frame**. Nothing in `engine/loom` carries reduced-motion handling of its own, so the caller has to gate the loop, and this file does.

### `backroom/demo.html` (228 lines) - the fixture
The real modules under a stubbed cashier. No shell, no bridge, no host, no network.

`?dark=1` (no `send`/`listen` at all), `?slow=1`, `?poor=1`, `?tier=1`, `?reduced=1`, `?lite=1`, and an unattended `?auto=` driver (`open` / `cage&n=N` / `esc` / `hold` / `drop`) that presses the same buttons a player would and writes what it saw into a log a headless dump can read back. Unhandled errors and rejections land in that log too, so **a throw is a failing test, not a line in a console nobody dumps**.

### Verification
A scratch runner drives Chrome headless against this page. **22 checks, all passing** on the tip of the stack.

Two live bugs the fixture caught are already fixed in the branches below it, which is the argument for it existing at all:

1. `countUp` painted a **minus balance** for a frame when the rAF stamp and `performance.now` disagreed (fixed in K1a).
2. The count-up's rAF loop **never terminated** under a frozen clock, and starved its own timer belt while doing it (fixed in K1a).

Shots (desktop 1600x900, phone 844x390 and 667x375, plus the paid cage and the dark cage) are attached in a comment.

### DIFF
--- a/ConditioningControlPanel/Resources/web/arcademy/backroom/backroom.css
+++ b/ConditioningControlPanel/Resources/web/arcademy/backroom/backroom.css
@@ -199,3 +199,47 @@ html.arc-mobile .bk-fly, html.ae-lite .bk-fly { box-shadow: none; }
  * to do is make sure nothing is left mid-flight looking broken. */
 html.arc-reduced .bk-bal.bk-thud { animation: none; }
 html.arc-reduced .bk-pot.bk-live .bk-pot-n { animation: none; opacity: 1; }
+
+/* ----------------------------------------------------------------------------
+ * THE HOLD RING (kit/holdring.js)
+ * The ring is a conic gradient driven by --bk-hold, so the sheet owns how the
+ * hold LOOKS and the module only ever owns how far along it is. Under the
+ * global reduced freeze the module steps the value in fifths instead of
+ * sweeping it, which is why there is no transition on it here: a transition
+ * would be flattened by the freeze and the ring would simply jump (trap 92).
+ * -------------------------------------------------------------------------- */
+.bk-hold { position: relative; --bk-hold: 0%; }
+.bk-hold::before {
+  content: ''; position: absolute; inset: -4px; border-radius: inherit; pointer-events: none;
+  background: conic-gradient(var(--bk-neon) var(--bk-hold), transparent var(--bk-hold));
+  -webkit-mask: radial-gradient(farthest-side, transparent calc(100% - 3px), #000 calc(100% - 3px));
+  mask: radial-gradient(farthest-side, transparent calc(100% - 3px), #000 calc(100% - 3px));
+  opacity: 0;
+}
+.bk-hold.bk-holding::before { opacity: 1; }
+.bk-hold.bk-held { box-shadow: 0 0 0 2px var(--bk-brass); }
+
+/* ----------------------------------------------------------------------------
+ * THE DROP (kit/dropbeat.js)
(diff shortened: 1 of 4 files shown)


CodeBambi:
## Shots

Taken headless against `backroom/demo.html` with the stubbed cashier, no host and no network. They live on the build machine at `C:\Users\PC\AppData\Local\Temp\claude\bk-k1-shots\`, because a PR comment cannot carry an image file from a CLI and none of these belong in the repo.

| File | What it shows |
|---|---|
| `floor-desktop-1600x900.png` | the floor at desktop width: header with balance and the lit exit sign, the pot, the cage, five cabinet frames |
| `floor-phone-844x390.png` | the same floor at 844x390, phone diet on |
| `floor-phone-667x375.png` | the narrow phone case, so the cage stepper and the cabinet column are checked at the tightest width the page supports |
| `cage-desktop-paid.png` | the cage after a confirmed exchange: banked chips, new balances, the warm receipt line and the tree trade-off |
| `floor-dark-cage.png` | `?dark`, the no-host case. Cage reads "Cash only. Come back when the line is up.", nothing throws, the floor is still the floor |
| `kit-bench-wheel-drop.png` | the fixture bench: the hold ring mid-press, the drop veil, and the wheel face with its Loom hub and brass pin |

To reproduce: serve `Resources/web/arcademy/` statically and open `backroom/demo.html`, optionally with `?dark`, `?poor`, `?tier`, `?reduced`, `?lite`, or `?auto=open|cage|esc|hold|drop|wheel|reel`.

CodeBambi:

PR #808, opened 2026-09-04
feat(arcademy): Back Room W1 - the door on the campus + chips at the counter

CodeBambi:
## What

Back Room W1, per `BACKROOM-CONTRACT.md` §3 and §4: the **door on the campus** and **chips at the Prize Counter**. Nothing behind the door yet.

1. **`shell/campus.js`** - `FACILITIES.backroom`, RM 005, the fifth window down the service alley, drawn through the same `facility()` helper as Records / Front Office / Prize Counter / The Locker. Compact plate, west door, neon sign reading BACK ROOM, the alley's twelve units of wall above it, a chip rack behind the glass. **Furniture, never a class**: no `ROOMS` row, no registry entry, no timetable involvement, nothing in `campusState`.
2. **`shell/shell.js`** - entering `backroom` walks the alley and then **dynamically imports `../backroom/index.js`** and calls `openBackRoom(ctx)`. Import failure raises the plan's dust sheet.
3. **`shell/prizecounter.js` + `.css`** - a third purse chip and a third shelf for `c`, read off the wallet echo.
4. **`core/lexicon.js` + `ArcademyHostService.cs`** - nine keys in both tables.

## Why the import is dynamic, and why the quad stays up

The casino wing is a floor of tables, a cage and a cabinet. A player who never opens that door should not pay for any of it on the slowest phone in the room, so the module is fetched on the click and cached once it lands. A host that does not ship the directory never resolves it, which is not an error, it is a school with one door still boarded up.

`showBackRoom()` is called at the end of the walk **with the campus still standing**, and it stays standing while the import is in the air. A failure raises the plan's own sealed card over a quad that never went anywhere, so there is nothing to "return" to and no empty screen to be stranded on. The dust sheet is `openFacilityCard({ sealed: true })` - the same card every other shut door on this campus already raises, reached through a new `campus.backroomDust()` seam because only the shell can know the import failed. The one path that *can* strand somebody (a module that lands and then refuses to mount) puts the player back on the board first.

A failure is **not** cached: a fetch that lost the network is worth trying again next time somebody tries the door.

## The ctx passed to `openBackRoom(ctx)`

The Annex's narrow surface plus the contract's two wire verbs. `backroom/index.js` is **not** in this PR (K1 owns that directory); this is the shape it will be handed.

| field | type | notes |
|---|---|---|
| `t(key, fallback)` | `(string, string) => string` | the lexicon, already merged with the mod skin |
| `lite` | `boolean` | the phone's cut-down flag |
| `reduced` | `boolean` | the motion setting |
| `log(line)` | `(string) => void` | the shell's console channel |
| `subject` | `object` | `init.subject`, read-only |
| `localDay` | `string` | `yyyy-mm-dd`, the LOCAL day the cage counts by |
| `gameName(key)` | `(string) => string` | a class's display name |
| `balance()` | `() => {t, k, c}` | the purse, host-owned |
| `earnedC()` | `() => number` | lifetime chips won, host-owned, off the same echo |
| `state()` | `() => object` | one page-owned blob in the meta store |
| `save(patch)` | `(object) => void` | merges into that one blob (headroom law) |
| `send(msg)` | `(object) => boolean` | see below |
| `listen(type, fn)` | `(string, fn) => () => void` | see below |
| `onExit()` | `() => void` | the way out, and the way out is the quad |

**`send(msg) -> boolean`** routes through the shell's existing `bridge.send` seam. Only `casino-request` and `triggers-request` go out; anything else returns `false` with a line in the log and no frame on the wire. The room never sees `bridge`.

**`listen(type, fn) -> unsubscribe`** subscribes through `bridge.on`. Only `casino-result` and `triggers-result` are accepted; anything else returns a no-op unsubscribe. Every live subscription is also tracked by the shell, so `clearScreen()` sweeps anything the room forgets to drop - the Back Room is the only screen on this floor holding a live host subscription, and a handler left on `casino-result` would paint a cage that has been thrown away.

The Esc ladder folds inward-out: `backRoomPage.escapeStep()` first, then out to the quad.

## Chips at the counter

- `SURFACES` is now `['t','k','c']`; chips come last because the Back Room is last in the alley.
- New exported **`curOf(row)`** replaces eight hand-written `row.cur === 'k' ? 'k' : 't'` ternaries. Those were fine at two surfaces and quietly wrong at three: a chip row would have been filed on the ticket shelf and priced with a stub. Unknown currencies still degrade to tickets, which is what all eight already assumed.
- The third purse chip is drawn **even at zero** - a purse that hides an empty currency teaches a player the currency does not exist.
- The chip shelf is only drawn when the catalog actually has `cur: 'c'` rows (the same nothing an empty token case has always drawn). **No chip rows exist in the vendored catalog yet**; D2 adds them. Verified here against a fixture.
- The chip glyph `.arc-chipmark` is **drawn**, not typed: a struck disc with an inset ring and a bite of edge spots, beside the ticket stub and the token coin in the shell's own primitives. A third unicode mark would be a third font risk on a page that fetches none (trap 2). Lavender rather than green because there is no green in the token plan, and the shelf's felt follows it: flat weave, no strip light, no glass.
- `HOLD_FROM_C = 2500` and `BEAT_DEAR_C = 500` (100 chips = 1 Sparkle, so 25 and 5 Sparkle). Worth re-auditing the day real chip prices land.
- Buying is unchanged: chip rows go out on the existing `prize-buy` send.

## Lexicon keys added (both tables)

`campus_room_backroom`, `campus_desc_backroom`, `backroom_sign`, `campus_backroom_status`, `backroom_dust`, `backroom_dust_line`, `wallet_chips`, `prize_shelf_chips`, `prize_shelf_chips_hint`.

All nine are in `core/lexicon.js` **and** `ArcademyHostService.cs` `NeutralLexicon` (a key the host does not ship falls back to English silently, and `MergeModTable` can only override a key the neutral table already knows). All under the 96-char mod cap.

## Verification

Headless-Chrome CDP rig against the page over a local static server, with a web shim standing in for the C# host (fixture catalog carrying three `cur: 'c'` rows, wallet `c: 2450 / earnedC: 18600`). **Measured, not eyeballed**: every shot dumps the Back Room's painted box, every piece of chrome that shares the plan with it, and the box intersection of each pair.

**15 shots, 9 viewports, zero clashes, zero page errors.**

Campus, `__RS.campus()` - the `clash` map is empty at every one of:

| viewport | Back Room box | clash |
|---|---|---|
| 390x844 portrait | 340,110 32x52 | none |
| 375x667 portrait | 311,107 27x44 | none |
| 360x640 portrait | 298,106 26x42 | none |
| 844x390 landscape | 682,341 52x31 | none |
| 1600x900 | 1400,788 120x71 | none |
| 1366x768 | 1195,672 102x61 | none |
| 1280x800 | 1173,700 107x63 | none |
| 1024x768 | clipped by the horizontal safe band, as all four alley windows are at 4:3 | none |
| 1920x1080 | 1680,945 144x85 | none |

Checked against the ID card, the post row, the top-right cluster, the desktop hint, the mail chip, the board and EMI. In portrait at 390x844 the room sits at x 340..372, y 110..162: the top-right cluster ends at y 94 (16px of air), the ID card starts at y 778, the mail chip at y 786, and the Prize Counter and Locker windows are the two windows above it in the same alley band.

**The 72 to 64 depth change was the rig's finding, not a taste.** A 72-deep window put its last two pixels under `.campus-hint` at 1366x768; the second pass caught the chip rack climbing into the RM plate. Both fixed and re-measured.

Door, `__RS.tryBackRoom()` at 390x844 and 1600x900 - tapping the window walks the alley, the shell's dynamic import fails on a host with no `backroom/` directory, and the plan raises: **THE BACK ROOM / NOT OPEN YET. / "Sheets over the tables and the lights off at the wall. Another night." / SEALED (disabled)**. `screen` never moved off the board.

Counter, `__RS.counter()` at 390x844 and 1600x900 - three purse chips (`5000 Tickets`, `500 Tokens`, `2450 Chips`, all three with their glyph), section order `pc-sect-t, pc-sect-k, pc-sect-c`, the third headed **The Back Room shelf** / "Chips only. What you carried out of the Back Room buys these.", and three chip-priced rows (400 / 1200 / 3200) each carrying `.arc-chipmark`.

Shots and per-shot logs: `C:/Users/PC/AppData/Local/Temp/claude/bk-w1-shots/` - `bk-{p390,p375,p360,l844,d1024,d1280,d1366,d1600,d1920}-campus.png`, `bk-{p390,d1600}-door.png`, `bk-{p390,d1600}-counter.png`, `bk-{p390,d1600}-chipshelf.png`.

`node --check` clean on all four JS files. No JS suite exists in this repo for the shell, so the fixture run above is the suite.

## Merge note

**Independent of the other Back Room PRs. The door shows a dust sheet until K1 lands.** Chip rows on the shelf show up when D2's catalog rows land; until then the shelf simply is not drawn and the purse reads a zero.

## Known gaps

- The campus top-right wallet chip and the Prize Counter booth's sill tray still show tickets and tokens only. Out of W1's scope; worth a follow-up once chips are actually earnable.
- `prize_counter_sub` still reads "Tickets on the shelf, tokens in the case". Leaving it: on the shipped catalog there are no chip rows yet, so a third clause would be the marquee lying.
- 597 changed lines (540 + 57), inside the ~600 cap.

### DIFF
--- a/ConditioningControlPanel/Resources/web/arcademy/core/lexicon.js
+++ b/ConditioningControlPanel/Resources/web/arcademy/core/lexicon.js
@@ -763,6 +763,34 @@ export const DEFAULT_LEXICON = Object.freeze({
   locker_open: 'Open Locker',
 
   /* ------------------------------------------------------------------------
+   * THE BACK ROOM (the casino wing, W1: the door and the chips).
+   *
+   * The fifth window in the service alley, and the only one that is not a
+   * counter. The voice drops a register again: the shop is warm, the locker is
+   * private, and this one is CALM - a house does not need to raise its voice.
+   * Every row is a NeutralLexicon candidate under the 96-character mod cap
+   * (trap 26), and the same rows are mirrored in the C# host or a host that
+   * ships without them falls back to English in silence (trap 123).
+   *
+   * The CHIP itself is a third currency and it sits with the other two: the
+   * counter reads `wallet.c` off the same echo the tickets and tokens ride.
+   * Chip PRICES and prize names are the host's, never this file's.
+   * ---------------------------------------------------------------------- */
+  campus_room_backroom: 'The Back Room',
+  campus_desc_backroom: 'Cash only. Chips only. The house always has time for you.',
+  backroom_sign: 'Back Room',
+  campus_backroom_status: 'Always open',
+  /* The dust sheet: the door is painted on the plan whether or not the room
+   * behind it has been built, because a landmark that comes and goes is a menu
+   * item. This is what it says on a host that has no room to open. */
+  backroom_dust: 'Not open yet.',
+  backroom_dust_line: 'Sheets over the tables and the lights off at the wall. Another night.',
+  /* the third shelf at the Prize Counter */
+  wallet_chips: 'Chips',
+  prize_shelf_chips: 'The Back Room shelf',
+  prize_shelf_chips_hint: 'Chips only. What you carried out of the Back Room buys these.',
+
+  /* ------------------------------------------------------------------------
    * THE PUBLIC ADDRESS SYSTEM (Counter Stock `pa_pack`, captions 2026-08-28).
    *
    * Thirty-six announcements. The recordings shipped first and were audio only;
--- a/ConditioningControlPanel/Resources/web/arcademy/shell/campus.js
+++ b/ConditioningControlPanel/Resources/web/arcademy/shell/campus.js
@@ -741,6 +741,27 @@ export const FACILITIES = Object.freeze({
   locker: Object.freeze({
     rect: [1260, 668, 108, 84], side: 'w', door: 710, stop: [1250, 710], rm: '004',
   }),
+  /* THE BACK ROOM (the casino wing, 2026-09-04). FIFTH WINDOW IN THE ALLEY,
+   * and it is the only one that is not a counter. Last door before the alley
+   * runs out of school: you walk past what you can buy and past what you
+   * already own, and then there is one more door. Same width, same west door,
+   * same three-step walk down the alley the other four take.
+   *
+   * TWELVE UNITS DOWN AND SIXTY-FOUR DEEP, not the ninety-six-by-eighty-four
+   * the pairs above it keep, and that is a MEASUREMENT rather than a taste:
+   * the plan's bottom band is where `.campus-hint` sits on a desktop (bottom
+   * 14px, the school's one line of instruction), and a fifth full-height
+   * window would have put this room's dressing under that sentence. The rig
+   * caught the last two pixels of a 72-deep window under the hint at 1366x768,
+   * so the window came in to 64 and the room now ends at plan y 828: eleven
+   * clear pixels at 1600x900 and just under five at 1366x768.
+   *
+   * FURNITURE, NEVER A CLASS. No ROOMS row, no registry, no timetable, nothing
+   * in campusState - the plan paints a door and the shell decides what is
+   * behind it. */
+  backroom: Object.freeze({
+    rect: [1260, 764, 108, 64], side: 'w', door: 796, stop: [1250, 796], rm: '005',
+  }),
 });
 
 /** ROOMS first, the two counters second. Pure; undefined for anything else. */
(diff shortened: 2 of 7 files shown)


CodeBambi:

PR #807, opened 2026-09-04
feat(arcademy): Back Room K1e - wheel face and trigger reels

CodeBambi:
Last of the five. The wheel face every spinning cabinet borrows, the reel builder Triple Trigger needs, and the one list of lexicon rows the whole stack owes the host.

**Merge order: #803 K1a, #804 K1b, #805 K1c, #806 K1d, then this one.** Each branch is stacked on the one before it, so merging out of order will show phantom conflicts. 473 lines here, 2550 across the stack.

## What is in it

`backroom/kit/loomwheel.js` (286) - the wheel face. One canvas, wedges drawn ONCE into an offscreen buffer and thereafter only rotated, because redrawing twenty-four labels a frame is how a wheel cooks a phone. The hub is a Loom spiral through `engine/loom/loomSpiral.js`, seeded so a given wheel always wears the same one. `spinTo(index, ms)` is told where to stop by the caller, which was told by the server: there is no rng in the landing, and the only random thing on the whole wheel is the hub's recipe. The pointer is fixed CSS furniture and the wheel turns under it, so a player can read the result at a glance. Settle is quintic out. Reduced motion crossfades to the answer over 240ms instead of travelling, `.ae-lite` gets a still hub, and a `setTimeout` ceiling at `dur + 400` means a stalled frame clock can never leave a wheel spinning with a stake on it.

`backroom/kit/triggers.js` (125) - the reels, and whose words are on them. `buildReel` keeps entries of exactly three words, dedupes case-insensitively, preserves the player's own order, caps at 12 and pads to 6 with glyph rows. Three words because a slot window is read in about a third of a second: two look like a label, four wrap.

`backroom.css` (+32) - the wheel block, plus fallbacks on the five kit-block token reads (`var(--bk-brass, var(--gold))` and so on). The tokens are declared on `.bk-room`, and a kit piece can be mounted anywhere, so the bench shot had an invisible pin until this landed.

`demo.html` (+33) - `?auto=wheel` and `?auto=reel` on the fixture bench.

## The trigger fallback does not read the mod, on purpose

[...]

PR #811, opened 2026-09-04
feat(arcademy): Back Room M4a - Scratcher

CodeBambi:
Machine 4 of the Back Room: the thumb cabinet. One free card a day from the
cashier, more for 50 chips, nine cells under a sheet of foil, and a prize when
any row comes up three of a kind.

**Merge order: K1 (#807), then this PR, then M4b (#818), then M4c (the
fixture).** Needs server S2 for live play; until S2 ships every press answers
`offline` and the cabinet says so warmly, which is a case the suite covers.

## What is in it

| File | lines | |
|---|---|---|
| `backroom/scratcher.js` | 447 | the cabinet: `{ key, title, mount, unmount }` |
| `backroom/scratcher.css` | 102 | its skin, lazy-linked on first open like `backroom.css` |

**549 lines, 2 files.** The bench that drives this machine is its own PR (M4c,
last in the stack) so that neither machine PR carries a harness.

## Why it is built this way

**The answer arrives before the scratch.** The server already knows what is on
the card the moment it takes the stake, so the module plays first and only then
lays the foil. A card that decided itself while the thumb was moving would be a
card with a thumb on it, and every honest scratcher ever printed was printed
before it was sold. It also means the whole receipt lands on ONE beat, when the
card has been read, rather than faking a stake leaving and a prize arriving as
two events the server never sent.

**The cleared fraction is an occupancy grid, not a pixel readback.** 18x18 cells,
one per 14 logical pixels, marked inside the brush radius on every move. Three
hundred and twenty four comparisons beat one `getImageData` by orders, and a
tainted or headless canvas can refuse the readback outright while the grid keeps
working. That is what lets the melt gate be asserted in a probe with no pixels.

**Reduced motion keeps the gesture and drops the melt.** A scratch is an act, not
an animation, so still mode still rubs; the foil simply goes at once instead of
dissolving. `.ae-lite` has no canvas at all: one tap-to-reveal plate, which is
the kindest thing to hand a thumb on a bus anyway. `liteMode()` reads BOTH
`kit.lite` and `html.ae-lite`, because a rig or a web host can arm one without
the other and the sheets have already put that phone on rations.

**Leaving mid-scratch banks the card.** Law VI: no key is bound here, the floor's
`escapeStep()` folds the cabinet, and `unmount()` settles an unread answer on the
way out - the ledger moved when the server answered, so the header may not go on
painting a number it has walked past.

**Odds come off the wire.** `kit.config().paytables.scratcher` renders the sheet;
the four-row table in the module is a floor for a status frame that has not
arrived and never a published number of its own (INPUT HONESTY).

## Answer fields consumed

`play('scratcher', stake, {})` where `stake` is **0 when
`free_scratch_available`** (the server decides what free means) else
`config.stakes.scratcher`, floor 50.

- `body.result.grid` - 9 x `{ kind:'spiral'|'word'|'chip', v:string }`. `spiral`
  draws a seeded Loom disc off `v`, once, at phase zero; the other two print `v`.
- `body.result.row` - `0..2` lights those three cells in brass, `-1` lights none.
- `body.result.prize` - the printed number and the bank; `>= 600` gets the `big`
  dealer bucket rather than `win`.
- `body.result.free` - `true` spends today's free card locally and flips the
  button to "50 chips".
- `body.chips` / `earnedC` / `pot` ride through `kit.settle(body, ticket)`.

Refusals: `insufficient_chips`, `busy`, anything else, `403` and transport
failure all map to a warm line that says nothing moved. No reason string is ever
shown raw.

## Lexicon keys (paste-ready)

```csharp
        // backroom/scratcher.js - THE SCRATCHER
        ["bk_sc_title"] = "The Scratcher",
        ["bk_sc_lede"] = "Three in a row pays. Rub anywhere.",
        ["bk_sc_buy_free"] = "FREE today",
        ["bk_sc_buy"] = "{0} chips",
        ["bk_sc_dealing"] = "Printing one.",
        ["bk_sc_rub"] = "Rub the foil.",
        ["bk_sc_tap"] = "tap to reveal",
        ["bk_sc_pct"] = "{0}% scratched",
        ["bk_sc_free_note"] = "On the house. One a day, every day.",
        ["bk_sc_win"] = "Three in a row. {0} chips.",
        ["bk_sc_none"] = "No row on that one. The foil was pretty though.",
        ["bk_sc_odds"] = "what a card can pay",
        ["bk_sc_odds_row"] = "{0} chips, one card in {1}",
        ["bk_sc_odds_rest"] = "anything else pays nothing. that is most cards.",
        ["bk_sc_poor"] = "Not enough chips for a card. The cage is just there.",
        ["bk_sc_offline"] = "The line to the counter is down. Nothing was charged.",
        ["bk_sc_busy"] = "The cashier is with someone. Give her a moment.",
        ["bk_sc_locked"] = "The bank does not serve this account yet. Nothing was charged.",
        ["bk_sc_refused"] = "The counter would not print that one. Nothing moved.",
```

Every row is under 96 characters so `MergeModTable` can re-voice it (trap 26).
`kit.t` falls back to BK_LEX and then to the KEY, and none of these rows are in
BK_LEX, so the module carries its own resolver: a host row wins, the table above
is the floor, and nothing can ever paint `bk_sc_win` on the ticket.

## Checks

The suite lives in M4c and needs all three PRs under it. Run against the M4c
tip, `bk-m4/run-checks.sh` drives `backroom/demo-machine.html` headless and
reads the dumped DOM: **43 of 43 pass, 23 of them this machine's.**

```
PASS sc: free card reads FREE today          PASS sc: no row lights nothing
PASS sc: a free card stakes nothing          PASS sc: no row gets sympathy copy
PASS sc: spent free card costs 50            PASS sc: a free card is spent once
PASS sc: a paid card stakes 50               PASS sc: lite gets a tap plate
PASS sc: nine cells off the answer           PASS sc: lite tap still reveals
PASS sc: foil is laid after the answer       PASS sc: still mode still pays
PASS sc: the foil melts at 60 percent        PASS sc: no chips refuses warmly
PASS sc: a quarter cleared does not          PASS sc: no chips prints no card
PASS sc: the winning row lights              PASS sc: offline refuses warmly
PASS sc: the prize is printed                PASS sc: the tier gate is standard
PASS sc: the prize banks                     PASS sc: leaving mid scratch is legal
                                             PASS sc: leaving still banks the card
```

The bench opens the cabinet by CLICKING ITS FRAME on the real floor, so what is
under test is index.js's own kit and mount contract rather than a hand-built
double that could drift from it. The scratch is real pointer events with real
`clientX/clientY` against a real `getBoundingClientRect`.

## Shots

On the build machine at `C:\Users\PC\AppData\Local\Temp\claude\bk-m4-shots\`,
because a PR body cannot carry an image from a CLI and neither belongs in the
repo.

| File | What it shows |
|---|---|
| `scratcher-desktop-1600x900.png` | free card played, foil melted, the brass row lit, the prize banked to 2,600, the odds sheet off the wire |
| `scratcher-phone-844x390.png` | the same at 844x390 with the phone diet on: no tilt, no paper shadow, the tap plate instead of the canvas |

Reproduce: serve `Resources/web/arcademy/` statically, then
`backroom/demo-machine.html?auto=full` plus any of `&free=0 &prize=0&row=-1
&poor=1 &tier=1 &dark=1 &reduced=1 &lite=1 &bare=1`.

## Notes for the reviewer

Nothing outside `backroom/` is touched, and nothing inside it that K1 shipped.

### DIFF
--- a/ConditioningControlPanel/Resources/web/arcademy/backroom/scratcher.css
+++ b/ConditioningControlPanel/Resources/web/arcademy/backroom/scratcher.css
@@ -1,0 +1,102 @@
+/* ============================================================================
+ * backroom/scratcher.css - the ticket.
+ *
+ * Lazy-linked by scratcher.js on first open, exactly like backroom.css. Every
+ * colour is a token off the shell or a mix of one, so a mod palette reskins the
+ * card for free. The literals are the PAPER and the FOIL, and they are literals
+ * for the same reason the felt is: both are remembered colours, and a violet
+ * scratch card reads as a bug rather than as a choice.
+ *
+ * TYPE: --disp for the ticket's signage, --mono for numbers, --body for copy.
+ * No serif (house camera law), no font host (trap 2).
+ *
+ * MOTION: the MELT is the only animation in here, the global freeze at the
+ * bottom of styles.css already flattens it, and scratcher.js does not wait on
+ * it under reduced motion either, so the card is read on every path. The diet
+ * drops the paper shadow and the tilt, which are the two things a phone pays a
+ * whole frame for.
+ * ==========================================================================*/
+
+.bk-sc { display: flex; flex-direction: column; gap: 12px; }
+.bk-sc-head { display: flex; align-items: baseline; gap: 10px; flex-wrap: wrap; }
+.bk-sc-head h3 { margin: 0; font-family: var(--disp); font-size: 15px; letter-spacing: .1em;
+  color: var(--bk-brass, var(--gold)); font-weight: 400; }
+.bk-sc-lede { font-size: 12px; color: var(--ink-faint); flex: 1 1 auto; }
+.bk-sc-stage { display: flex; gap: 18px; align-items: flex-start; flex-wrap: wrap; }
+
+/* ---------------------------------------------------------------- the ticket */
+.bk-sc-ticket {
+  flex: 0 0 auto; width: 288px; padding: 12px; border-radius: 10px;
+  background: linear-gradient(180deg, #efe3c8, #ddceac); color: #3a2c14;
+  box-shadow: 0 10px 26px rgba(0, 0, 0, .45); transform: rotate(-1.2deg);
+}
+.bk-sc-thead, .bk-sc-tfoot { display: flex; justify-content: space-between; gap: 8px; align-items: baseline; }
+.bk-sc-thead { font-family: var(--disp); font-size: 11px; letter-spacing: .1em; margin-bottom: 8px; }
+.bk-sc-tfoot { font-family: var(--mono); font-size: 10px; color: #7a6440; margin-top: 8px; }
+.bk-sc-pct { white-space: nowrap; }
+
+/* The window is the hole in the paper. The foil sits OVER it, never beside it. */
+.bk-sc-window { position: relative; width: 100%; aspect-ratio: 1 / 1;
+  border-radius: 8px; overflow: hidden; background: #1a1030; }
+.bk-sc-grid { position: absolute; inset: 0; display: grid; gap: 4px; padding: 4px;
+  grid-template-columns: repeat(3, 1fr); grid-template-rows: repeat(3, 1fr); }
+.bk-sc-cell {
+  display: grid; place-items: center; border-radius: 6px; padding: 3px; overflow: hidden;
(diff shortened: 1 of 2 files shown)


CodeBambi:

PR #813, opened 2026-09-04
feat(arcademy): Back Room M1b - Twenty-One skin and bench

CodeBambi:
## What

The second half of Back Room M1, split off so each PR stays under the 600 line cap. Based on `feat/backroom-m1-twentyone` (M1a, the module). Two new files, both in `backroom/`:

- `twentyone.css` (170) - the table's skin, lazy-linked by the module.
- `twentyone.demo.html` (268) - the bench that deals canned hands at the real module.

## Why

**The skin borrows, it never redeclares.** `--bk-felt`, `--bk-neon`, `--bk-neon-2`, `--bk-brass`, `--bk-edge` and `--bk-plate` are declared on `.bk-room` in `backroom.css`; every read here carries a fallback, because a machine can be mounted on a bench with no room around it and an unfallen `var()` paints furniture nobody can see. Display face for the name, mono for numbers, body for her sentences, no serif anywhere.

The phone rules are not decoration. A flex item's default `min-width` is `auto`, so the odds sentence refused to shrink and pushed the whole page sideways at 390; `min-width: 0` on the hud children lets it wrap instead, the odds line takes its own row under the diet, and the way out stops being pushed to the far edge where a thumb cannot reach it. The felt is capped at 700px so it reads as furniture rather than as a wall.

**The bench mounts the real module** under a stubbed `kit`, so the tests exercise the shipping code and not a copy of it. `?auto=` plays one scripted script and prints what landed into a log div the runner reads:

`deal, bust, win, push, bj, double, poor, offline, resume, lock, unmount, open, floor`

plus `?dark=1` (no line), `?reduced=1`, `?lite=1`, `?slow=1` (hold the answer so the lock is observable), and `?w=` which caps the bench width. `?w=` exists because headless Chrome on Windows will not open a window narrower than about 500 CSS px; nothing in the table's own CSS reads `vw` or a media query, so a capped bench is the same layout a phone gets. `?auto=floor` imports the real `./index.js` and drives the cabinet through the actual floor.

`demo.html` in this folder set the precedent for a fixture living in the repo.

## Tests

Runner lives outside the repo (`scratchpad/bk-m1/run-checks.sh`, K1 harness shape, headless Chrome against a local static server):

```
PASS table opens, odds off config          PASS blackjack rings the jackpot
PASS stake label off config                PASS DOUBLE DOWN shows and pays
PASS no line: the table reads dark         PASS double is offered on the deal
PASS deal sends the right frame            PASS no chips reads warmly
PASS deal paints both hands                PASS offline says nothing was taken
PASS hole card hides the count             PASS an open hand resumes at mount
PASS deal settles the balance              PASS resumed hand shows the verbs
PASS DEEPER into a bust reads warm         PASS every verb locks mid answer
PASS bust reveals the hole card            PASS a second press cannot stake
PASS HOLD into a win pays                  PASS unmount mid hand is silent
PASS a win banks out of the felt           PASS still mode deals the same hand
PASS a push comes home untouched           PASS phone diet deals the same hand
PASS blackjack is a beat                   PASS the floor mounts the cabinet
PASS the cabinet is no dust sheet          PASS the ladder folds it back home
PASS no sideways scroll at 390             PASS no sideways scroll at 360

---- 30 passed, 0 failed
```

## Shots

Off this bench, in `C:/Users/PC/AppData/Local/Temp/claude/bk-m1-shots/`:

- `table-desktop-hand.png`, `table-desktop-blackjack.png`
- `table-phone-390-lite.png`, `table-phone-360-win.png`

## Merge note

**Stacked on K1 (#807); needs server S2 for live play.** Merge this in the same wave as M1a, which is unskinned without it.

### DIFF
--- a/ConditioningControlPanel/Resources/web/arcademy/backroom/twentyone.css
+++ b/ConditioningControlPanel/Resources/web/arcademy/backroom/twentyone.css
@@ -1,0 +1,170 @@
+/* ============================================================================
+ * backroom/twentyone.css - THE TABLE'S SKIN.
+ *
+ * Lazy-linked by twentyone.js the first time the cabinet is opened, the same
+ * way backroom/index.js links the floor's sheet, so a player who never walks
+ * to this table never pays for it.
+ *
+ * IT BORROWS, IT DOES NOT REDECLARE. --bk-felt, --bk-neon, --bk-brass and the
+ * rest are declared on .bk-room in backroom.css, and every read here carries a
+ * fallback because a machine can be mounted on a bench with no room around it
+ * (the K1 lesson: an unfallen var() paints furniture nobody can see).
+ *
+ * NO SERIF ANYWHERE (house camera law). Display for the name, mono for
+ * numbers, body for the dealer's sentences, and nothing else.
+ *
+ * The still-mode and phone blocks sit at the bottom. The global freeze at the
+ * end of styles.css already flattens CSS animation under html.arc-reduced; the
+ * blocks here are for the things a flattened duration still leaves looking
+ * wrong, and for the glow a phone should not be paying for.
+ * ==========================================================================*/
+
+.bk21 { display: flex; flex-direction: column; gap: 10px; }
+
+/* ------------------------------------------------------------------- the hud */
+/* min-width:0 on every child is not tidiness, it is the fix: a flex item's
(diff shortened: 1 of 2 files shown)


CodeBambi:

PR #814, opened 2026-09-04
feat(arcademy): Back Room M2a - Triple Trigger reel window and skin

CodeBambi:
## What

The mechanical half of TRIPLE TRIGGER, the Back Room slot machine whose reels are the player's own three word triggers. This part is the reel window and the skin only. The cabinet that drives it is M2b, stacked on this.

**`backroom/triple-reel.js`** (309) exports `createReelWindow(opts)` and `REELS`. It owns the three strips, the roll and the settle, and nothing else.

**`backroom/triple.css`** (202) is the skin, in the backroom palette. No serif.

## Why it is split

M2 as one file was 932 lines. The reel is transform maths and the cabinet is copy and rules, and they were not helping each other read. Splitting on that seam puts this part at 511 changed lines and the cabinet at 592, both under the cap.

## The parts that are load bearing

**The cell height is a custom property.** A reel parks at `translateY(calc(var(--bk-tt-cell) * -n))`, so the module never measures a pixel and the phone diet can shorten the cells (54px to 40px) without a reel missing the payline.

**The roll is a rAF loop and it gates itself.** The global freeze at the bottom of `styles.css` kills animation and transition but cannot reach a rAF (trap 92), so `startRoll` returns before the first frame under reduced motion and the still mode path is an opacity class the module times instead.

**The blank has somewhere to land.** Every strip carries one extra glyph cell past the end of the list, and `posFor()` maps the server's `-1` onto it. The reels are three copies of the list so a roll never runs off the top.

**The landed index is in the DOM.** `land()` writes `dataset.at`, which is how the M2c fixture asserts "the reels settled where the server said" out of a dumped DOM with no instrumentation inside the module.

**A held reel gets no thud.** `land()` skips its sfx if the reel never turned. You paid five chips for it to stay still and it should sound like it stayed still.

**A roll cannot spin forever.** `ROLL_CEILING_MS` is armed before the first frame, and the cabinet's `onCeiling` unlocks the buttons and says so. A machine that has taken your chips and gone quiet is the worst thing this room could do.

**No filter on the smear.** The rolling look is `opacity: .78`, not a blur. A `filter` on a layer that is transformed every frame repaints the whole layer every frame (trap 36) and a phone pays that first.

## Merge note

Stacked on K1 (#807). Nothing imports this yet, so it is inert until M2b lands. Needs server S2 for live play.

### DIFF
--- a/ConditioningControlPanel/Resources/web/arcademy/backroom/triple-reel.js
+++ b/ConditioningControlPanel/Resources/web/arcademy/backroom/triple-reel.js
@@ -1,0 +1,309 @@
+/* ============================================================================
+ * backroom/triple-reel.js - the three reels, and nothing else.
+ *
+ * The window of TRIPLE TRIGGER: three columns of words that turn, and stop
+ * where they are told to stop. It knows nothing about chips, stakes, holds as
+ * a PRICE, odds or the dealer. Hand it a list of trigger rows and it paints
+ * three reels; hand it three indices and it lands on them. That is the whole
+ * surface, and keeping it that small is what makes the cabinet on top of it
+ * readable and this half testable on a bench with no cashier anywhere near it.
+ *
+ * FOUR THINGS THIS FILE IS BUILT AROUND.
+ *
+ *  - THE REELS ARE THEATRE (LEDGER TRUTH). The server drew the outcome before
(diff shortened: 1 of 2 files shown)


CodeBambi:

PR #815, opened 2026-09-04
feat(arcademy): Back Room M2 - Triple Trigger (cabinet)

CodeBambi:
## What

The TRIPLE TRIGGER cabinet. `backroom/triple.js` (592) exports `{ key:'triple', title, mount(root, kit, ctx), unmount() }` and the floor picks it up through the existing `probeCabinet` import, so no file outside `backroom/triple*` is touched.

Reel k shows word k of every trigger row the player owns, built with `reelRows`/`buildReel` on mount, `n` = the list length. A win is the player's own sentence read back at them: a ROYAL reads the whole trigger across the three reels, a SCRAMBLE reads the new sentence the three reels just made, a PAIR is a small warm win, and a blank sometimes gets a word from the dealer and sometimes does not.

## Why it is built this way

**LEDGER TRUTH.** The cabinet computes no chips. It sends a stake and renders what comes back, and the header only ever moves through `kit.settle(body, bankFrom)`.

**INPUT HONESTY.** The odds sheet is rendered from `kit.config()`, all three rows by hold count, with the row for the current hold count marked. It says out loud that the multipliers are on the base stake and that a hold is never multiplied. `readPay()` reads the paytable loosely (a `rows[]` or flat keys, a probability or a denominator) so a shape change on the server side degrades to a sheet rather than to a wrong sheet.

**EXITS SACRED.** No key is bound anywhere in this module and no dialog is opened. "back to the floor" is built first and is never disabled, and `unmount()` is safe with a stake in flight and the reels turning: every timer is cleared, the reel, the ring and the drop beat are destroyed, and a late answer lands on a dead instance without throwing.

**LOSSES GET SYMPATHY, and not every time.** A dealer who commiserates after every single blank pull is a dealer who is managing you, so the loss line is a coin flip weighted at 0.45. Every refusal reads as a person declining rather than a machine erroring.

**The holds are dead until the first pull.** Selling five chips to hold a reel that has not been anywhere yet is selling nothing. The price line says why instead of the buttons just going grey.

## Why the royal does not speak out loud

[...]

PR #816, opened 2026-09-04
test(arcademy): Back Room M2 - Triple Trigger fixture

CodeBambi:
## What

`backroom/triple-demo.html` (264), the fixture for the TRIPLE TRIGGER cabinet.

It opens the **real** room with `openBackRoom(ctx)`, clicks the **real** Triple frame on the floor and holds the **real** lever for its full 900ms. The only stub is the cashier. There is no runner and no framework in here on purpose: the fixture is the harness, so it cannot drift from it, which is how the K1 kit is tested and this stays in step with it.

## How it asserts

The cashier answers `casino-request` and `triggers-request` the way the host will, out of a canned table with one row per outcome class:

```js
none:     { reels: [0, 1, -1], line: 'none',     mult: 0,   potWon: 0,     payout: 0 }
pair:     { reels: [2, 2, 0],  line: 'pair',     mult: 1.2, potWon: 0,     payout: 12 }
scramble: { reels: [0, 1, 2],  line: 'scramble', mult: 3,   potWon: 0,     payout: 30 }
royal:    { reels: [1, 1, 1],  line: 'royal',    mult: 100, potWon: 41800, payout: 1000 }
```

Each row names the reel indices, so "did the reels settle where the server said" is a question a **dumped DOM can answer by itself**: the module writes the landed index to `dataset.at` and the probe reads it out of the markup. The whole outbound frame is logged too, because on this machine the stake and the holds are the assertion that matters most.

The trigger list it hands over is deliberately awkward: one entry is two words so the kit's three word filter is exercised, one duplicates an authored row so the dedupe is, and the survivors are short enough to force a padded row out of the authored set.

Query params: `?line=`, `?poor`, `?tier`, `?dark`, `?reduced`, `?lite`, `?shot`, `?auto=open|pull|holds|unmount`. `window.onerror` and `unhandledrejection` both write `THREW`, so a throw fails a check even when the assertion would otherwise have passed.

## Test output

Runner is `scratchpad/bk-m2/run-checks.sh` (static `python -m http.server` plus `chrome --headless=new --dump-dom`, the K1 harness style).

```
PASS cabinet mounts off the floor
PASS glyph row renders as a spiral
PASS the player's words are reel one
PASS price line, base and the hold
PASS odds sheet from config
PASS holds dead before the first pull
PASS lever live once words arrive
PASS no line: reels land as told
PASS no line: warm, never a jab
PASS no line: the stake is taken
PASS pair: reels land as told
PASS pair: pays and says so
PASS pair: banked to the header
PASS scramble: three different rows
PASS scramble: reads the new sentence
PASS royal: three on one trigger
PASS royal: reads the trigger back
PASS royal: the pot comes with it
PASS royal: banked, pot and all
PASS wire body: stake 10, no holds
PASS wire body: n is the list length
PASS holds open after a pull
PASS two holds move the price to 20
PASS two holds ride the wire
PASS held reels are on the wire
PASS insufficient chips reads warmly
PASS insufficient: nothing was taken
PASS tier gate, the standard copy
PASS offline: the cabinet is honest
PASS offline: the lever is dead
PASS still mode lands the reels
PASS still mode still reads the royal
PASS phone diet: tap lever pulls
PASS phone diet: pays the same
PASS unmount mid spin folds it
PASS unmount mid spin: back on floor
PASS unmount mid spin: stays folded
---- 37 passed, 0 failed
```

What the wire looks like on a two hold pull, straight out of the log:

```
{"op":"play","body":{"machine":"triple","stake":20,"playId":"...","input":{"holds":[true,false,true],"n":6}}
```

The unmount check is the one worth reading twice: it holds the lever, waits 120ms so the reels are turning and a stake is on the wire, sends Escape through the shell's ladder, and then asserts the cabinet is folded, the floor is back, **and** that it is still folded 2.6s later once the canned answer has arrived at a dead instance.

## Shots

Under `C:/Users/PC/AppData/Local/Temp/claude/bk-m2-shots/`:

- `m2-desktop-1600x900.png` - a royal. DROP FOR ME across the payline, the window lit gold, "ROYAL. DROP FOR ME. Paid 1,000 chips. And the pot came with it. 41,800 chips.", the pot back to 0 and the header at 2,240. The odds sheet beside it with the hold 0 row marked.
- `m2-phone-844x390.png` - the diet. Tap lever, a scramble reading GOOD FOR WARM, the glyph blank visible on reel one.
- `m2-phone-667x375.png` - the diet at the narrow gate. One column, the sheet below the machine, a pair paying 12.

## Merge note

Stacked on M2b (#815) and M2a (#814), which are stacked on K1 (#807). Needs server S2 for live play. Merge in order: #814, #815, then this.

### DIFF
--- a/ConditioningControlPanel/Resources/web/arcademy/backroom/triple-demo.html
+++ b/ConditioningControlPanel/Resources/web/arcademy/backroom/triple-demo.html
@@ -1,0 +1,264 @@
+<!doctype html>
+<html lang="en">
+<head>
+<meta charset="utf-8">
+<meta name="viewport" content="width=device-width,initial-scale=1">
+<title>Triple Trigger - cabinet fixture</title>
+<!-- The REAL floor and the REAL cabinet under a STUBBED cashier, exactly the
+     way backroom/demo.html does it: the room is opened, the Triple Trigger
+     frame is CLICKED, and everything after that is the shipped module. There
+     is no test runner and no framework here on purpose. The fixture IS the
+     harness, so it can never drift from it.
+
+     ?line=none|pair|scramble|royal   which answer the cashier gives back
+     ?poor=1     the cashier refuses for want of chips
+     ?tier=1     the cashier refuses with the tier gate (403)
+     ?dark=1     no send/listen at all: the cabinet must draw itself unplugged
+     ?reduced=1  arms html.arc-reduced, the global still mode
+     ?lite=1     arms html.ae-lite + html.arc-mobile, the phone diet (tap lever)
+     ?auto=open      mounts the cabinet and reports what it painted
+     ?auto=pull      pulls the lever once and reports where the reels landed
+     ?auto=holds     pulls, holds two reels, pulls again, reports the wire body
+     ?auto=unmount   pulls and steps back to the floor MID SPIN
+     ?shot=1     frames the cabinet for a screenshot (hides the fixture chrome)
+     Everything is written into #log, which is what the headless probe reads
+     back out of the dumped DOM. -->
+<link rel="stylesheet" href="../styles.css">
+<style>
+  body { margin: 0; padding: 16px 14px 60px; background: var(--ground); color: var(--ink);
+         font-family: var(--body); }
+  .wrap { max-width: 1100px; margin: 0 auto; }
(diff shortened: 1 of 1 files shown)


CodeBambi:

PR #818, opened 2026-09-04
feat(arcademy): Back Room M4b - Prize Wheel

CodeBambi:
Machine 5 of the Back Room: the loudest thing on the floor. Five hundred chips
buys one spin at seven wedges, two of them worth waiting for.

**Merge order: K1 (#807), then M4a (#811), then this PR, then M4c (the
fixture).** Needs server S2 for live play.

## What is in it

| File | lines | |
|---|---|---|
| `backroom/wheel.js` | 358 | the cabinet: `{ key, title, mount, unmount }` |
| `backroom/wheel.css` | 72 | the furniture around the face |

**430 lines, 2 files.** The bench that drives both machines is its own PR (M4c,
last in the stack) so that neither machine PR carries a harness.

The FACE has no skin in here on purpose. `.bk-wheel`, its pin and its fade are
already in `backroom.css` because the face is kit and the kit's skin belongs
with the kit; `wheel.css` is only the button, the prize card and the table.

## The wedge layout, and why there are fifty slices

`kit/loomwheel.js` draws every slice the same size (`const span = TAU /
wedges.length`). Hand it the seven wedges and a two-percent wedge and a
thirty-six-percent wedge come out as identical pie slices, which is a lie a
player cannot even see being told. So the face is handed **fifty slices**, one
per two percent, in seven contiguous runs, and **only the middle slice of a run
carries a label** (`buildFace` skips an empty one). Every wedge's arc is then
exactly its published weight, and the gold rule the face strokes between slices
becomes a countable odds mark: **one tick is two spins in a hundred**. INPUT
HONESTY costs forty-three extra slices and buys a face a player can audit at a
glance.

| wedge | face label | weight /100 | slices | pin lands on slice | kind | sku |
|---|---|---|---|---|---|---|
| 0 | 250 chips | 36 | 0-17 | 8 | chips | |
| 1 | 500 chips | 20 | 18-27 | 22 | chips | |
| 2 | Scratch card | 18 | 28-36 | 32 | item | `bk_scratcher` |
| 3 | Insurance | 12 | 37-42 | 39 | item | `bk_insurance` |
| 4 | 1,000 chips | 8 | 43-46 | 44 | chips | |
| 5 | Late Slip | 4 | 47-48 | 47 | item | `late_slip` |
| 6 | Dealer's Visor | 2 | 49 | 49 | item | `bk_visor` |

Fifty slices, sum 100. `spinTo` is given the run's MIDDLE slice so the wedge
[...]

PR #819, opened 2026-09-04
test(arcademy): Back Room M4c - Scratcher + Prize Wheel fixture

CodeBambi:
The machine bench for the two M4 cabinets, split out so that neither machine PR
has to carry its own harness.

**Merge order: K1 (#807), then M4a (#811), then M4b (#818), then this PR.**

**This PR's suite only passes once all three land under it.** On its own the
file is a page that opens two cabinets which are not there yet, and the floor
draws them as dust sheets, which is exactly what `index.js` promises for a
missing module. Nothing here is worth running until M4a and M4b are merged.

## What is in it

| File | lines | |
|---|---|---|
| `backroom/demo-machine.html` | 287 | the bench: real floor, pretend cashier, unattended driver |

**287 lines, 1 file.** Nothing outside `backroom/` is touched, and nothing
inside it that K1 shipped: `demo.html` is K1's kit bench and is left alone, so
the two benches can drift apart without either breaking the other.

## Why it is built this way

**It clicks a cabinet frame.** The bench calls the real `openBackRoom(ctx)` and
then finds the cabinet by its own printed name and `.click()`s it, exactly as a
player does. What is under test is therefore `index.js`'s own kit and its mount
contract, not a hand-built double that could quietly drift from the kit the
machines will actually be handed in production. `probeCabinet`, the dust sheet
fallback, `escapeStep`, `kit.settle` and the balance chip are all live.

**The gestures are real.** The scratch dispatches real `PointerEvent`s with real
`clientX/clientY` against a real `getBoundingClientRect`, so the occupancy grid
and the 60 percent melt gate are exercised the way a thumb exercises them. The
spin is the real `createLoomWheel` face landing on the cashier's wedge, ceiling
and all.

**The cashier holds a purse.** 2,000 chips, and a stake visibly leaves it and a
prize visibly arrives, because a fixture whose balance never moves cannot catch
a machine that forgot to bank. It is the only thing in the file that knows a
number.

**It writes down what it saw.** `?auto=` drives the page unattended and appends
plain lines to `#log`, which is what the headless probe reads back out of
`--dump-dom`. That is the whole reason there is no framework here: the fixture
IS the harness, so it can never drift from it.

**The Esc ladder is held the way the shell holds it.** One document listener
that tries `room.escapeStep()` first and closes the room second, so a machine
that illegally bound a key would show up as a cabinet that will not fold
(Law VI).

## The flags

| flag | what it picks |
|---|---|
| `?m=scratcher\|wheel` | which cabinet frame to click |
| `?dark=1` | no `send`/`listen` at all: every press must refuse warmly |
| `?poor=1` / `?tier=1` | `insufficient_chips` / the 403 `arcademy_locked` gate |
| `?free=0` | today's free card is already spent |
| `?prize=N` `?row=N` | the scratcher's prize and winning row (`-1` = none) |
| `?wedge=0..6` | the wheel's landing |
| `?fallback=1` | an item wedge finds the shelf full, so the house pays 250 |
| `?reduced=1` `?lite=1` | `html.arc-reduced` / `html.ae-lite` + `arc-mobile` |
| `?bare=1` | folds the cage and the pot away so a shot is of the CABINET |
| `?auto=open\|full\|part\|leave\|refuse` | the unattended cases |

## Checks

`bk-m4/run-checks.sh` (scratch, not in the repo) points headless Chrome at this
page once per case and greps the dumped `#log`. **43 passed, 0 failed** at this
branch's tip.

```
== scratcher ==
PASS sc: free card reads FREE today          PASS sc: no row lights nothing
PASS sc: a free card stakes nothing          PASS sc: no row gets sympathy copy
PASS sc: spent free card costs 50            PASS sc: a free card is spent once
PASS sc: a paid card stakes 50               PASS sc: lite gets a tap plate
PASS sc: nine cells off the answer           PASS sc: lite tap still reveals
PASS sc: foil is laid after the answer       PASS sc: still mode still pays
PASS sc: the foil melts at 60 percent        PASS sc: no chips refuses warmly
PASS sc: a quarter cleared does not          PASS sc: no chips prints no card
PASS sc: the winning row lights              PASS sc: offline refuses warmly
PASS sc: the prize is printed                PASS sc: the tier gate is standard
PASS sc: the prize banks                     PASS sc: leaving mid scratch is legal
                                             PASS sc: leaving still banks the card
== prize wheel ==
PASS pw: the frame opens                     PASS pw: consumables land too
PASS pw: a spin stakes 500                   PASS pw: still mode still lands
PASS pw: it plays the wheel                  PASS pw: lite still lands
PASS pw: seven wedges, honest widths         PASS pw: the stake is on the button
PASS pw: it lands where told                 PASS pw: no chips refuses warmly
PASS pw: a chip wedge banks                  PASS pw: offline refuses warmly
PASS pw: a chip wedge names the chips        PASS pw: the tier gate is standard
PASS pw: an item wedge names the item        PASS pw: leaving mid spin is legal
PASS pw: an item says where it went          PASS pw: leaving still banks the spin
PASS pw: a full shelf says so honestly       PASS pw: the odds table is published
```

Reproduce: serve `Resources/web/arcademy/` statically, then open
`backroom/demo-machine.html` with any combination of the flags above.

## Shots

Taken through this bench, on the build machine at
`C:\Users\PC\AppData\Local\Temp\claude\bk-m4-shots\`: `scratcher-desktop-1600x900.png`,
`scratcher-phone-844x390.png`, `wheel-desktop-1600x900.png`,
`wheel-phone-844x390.png`. A PR body cannot carry an image from a CLI and
neither belongs in the repo.

### DIFF
--- a/ConditioningControlPanel/Resources/web/arcademy/backroom/demo-machine.html
+++ b/ConditioningControlPanel/Resources/web/arcademy/backroom/demo-machine.html
@@ -1,0 +1,287 @@
+<!DOCTYPE html>
+<html lang="en"><head>
+<meta charset="utf-8">
+<meta name="viewport" content="width=device-width,initial-scale=1">
+<title>The Back Room - machine fixture</title>
+<!-- The REAL floor with a REAL machine mounted on it over a pretend cashier at
+     the bottom of this file. No shell, no bridge, no host, no ledger. The
+     cabinet is opened by CLICKING ITS FRAME, exactly as a player opens it, so
+     what is under test is the whole mount contract rather than a hand-built kit
+     that could quietly drift from index.js's.
+
+     ?m=scratcher|wheel  which frame to click (default scratcher)
+     ?dark=1     no send/listen at all. Every press must refuse warmly.
+     ?poor=1     the cashier refuses for want of chips. ?tier=1 the 403 gate.
+     ?free=0     today's free card is already spent.
+     ?prize=N    the scratcher's prize, ?row=N its winning row (-1 = none).
+     ?wedge=N    the wheel's landing, 0..6. ?fallback=1 makes an item wedge
+                 find the shelf already full, so the house pays 250 instead.
+     ?reduced=1  arms html.arc-reduced. ?lite=1 arms html.ae-lite + arc-mobile.
+     ?bare=1     folds the cage and the pot away, so a 1600x900 shot is of the
+                 CABINET rather than of the furniture above it.
+     ?auto=...   drives it unattended and writes what it saw into #log, which is
+                 what the headless probe reads back out of the dumped DOM.
+     No framework here on purpose: the fixture IS the harness, so it can never
+     drift from it. -->
+<link rel="stylesheet" href="../styles.css">
+<style>
+  body { margin: 0; padding: 14px 12px 48px; background: var(--ground); color: var(--ink);
+         font-family: var(--body); }
+  .wrap { max-width: 1100px; margin: 0 auto; }
+  header.fx { display: flex; align-items: baseline; gap: 12px; flex-wrap: wrap;
+              border-bottom: 3px solid var(--pink); padding-bottom: 8px; margin-bottom: 12px; }
+  header.fx h1 { font-family: var(--disp); font-size: 18px; margin: 0; letter-spacing: .08em; }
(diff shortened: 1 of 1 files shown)


CodeBambi:

PR #820, opened 2026-09-04
feat(arcademy): Back Room INTEGRATION - the five cabinets on one floor

CodeBambi:
The Back Room, assembled. This branch merges the whole wave 1 desktop stack into one tree,
fixes what only showed up once the pieces were side by side, and mirrors the wing's words into
the host lexicon. Every harness in the wave is green on the merged tree.

## Merge order (all `--no-ff`, no conflicts in any of the eight)

1. `feat/backroom-d1-cage` (#797) - chips shape, casino bridge, cage adoption
2. `feat/backroom-d2-catalog` - the chip shelf, catalog rows
3. `feat/backroom-w1-door` (#808) - the door on the campus, chips at the counter
4. `feat/backroom-k1e` (#807) - the kit, the floor and THE CAGE
5. `feat/backroom-m3-spiral-skin` (#810) - The Spiral
6. `feat/backroom-m1b-twentyone-skin` (#813) - Twenty-One
7. `feat/backroom-m2-triple-test` (#816) - Triple Trigger
8. `feat/backroom-m4c-fixture` (#819) - The Scratcher and The Prize Wheel

Not a chain: D1, D2 and W1 sit on `main`, and M1 to M4 are four SIBLINGS off K1e. There is
therefore no base that shows the integration work alone. This PR is based on the last branch in
the order (#819) because that hides the most, but D1, D2, W1, M1, M2 and M3 will still show in
the diff and are reviewed in their own PRs. Merge them in the order above, then this.

## What the assembly turned up, and what it cost

**One `CurChips`, not two.** D1 and D2 both landed the currency constant. Kept one.

**The floor is no longer under sheets.** All five machine modules exist now, so `index.js`
imports five and paints none of them as a dust sheet. Nothing needed registering: the floor
probes `./<key>.js` on the click and each machine exports the `{ key, title, mount, unmount }`
the contract asks for. The K1 harness assertion that counted 5 sheets now counts 0.

**The shell's `reduced` spelling.** `shell.js` `backRoomCaps()` hands the scene `reduced`, and
K1's `stillMode()` only read `motion.reducedMotion`, so reduced motion arrived late (once the
`arc-reduced` class landed) rather than on the first paint. Reads both now.

**`bank()` mixed clocks.** `kit/chips.js` `bank()` started on `performance.now` and measured with
the rAF stamp, which is the exact bug `countUp()` one function above was written to fix. Both
now read one `clock()` and both carry the frame floor, so a headless probe on a virtual-time
[...]

PR #823, opened 2026-09-04
feat(arcademy): Back Room G1 - no door until the house says open

CodeBambi:
## The rule

**No door until the house says open.** The Back Room window in the alley is drawn only when `GET /v2/arcademy/casino` (server PR CC-Labs-llc/CCP-Server#139) answers `open: true` for this player. The shell asks once per campus build through the existing `casino-request` frame with its own `reqId`, throttled to one ask per five minutes, with a four second timeout. The campus builds first and the door is lit when the answer lands, so a slow server never holds the quad.

Anything else leaves the alley as it was before the wing, four windows and no hole: offline (status 0), a 403, a timeout, a server that has no `open` field yet, or `open: false`.

**No host change.** Desktop (D1) and the web relay already forward `casino-request` and answer `casino-result`. The shell's ask filters on its own `reqId`, so it can never eat a receipt the room is waiting on.

## Mid-session close

If the flag closes while a player is at a table, the next cage or play answer is `403 casino_closed`. All six refusal maps (cage, Spiral, Triple Trigger, Twenty-One, Scratcher, Prize Wheel) now check that reason **before** the 403 branch and print `bk_closed`: "The house is closed tonight. Your chips keep. Come back when the sign is lit." The exit stays lit as always. Without the reorder the 403 branch would have printed "the bank does not serve this account", which is the wrong story. `bk_closed` is mirrored into `NeutralLexicon`.

`showBackRoom` also refuses with the plan's sealed card when the door is not lit, for a keybind or stale handle that asks anyway.

## Harness

Copied shots rig (scratchpad `bk-g1/shots`, the shim gains `?bk=open|closed|silent`), run headless through the real shell on this branch:

| case | door | campus | result |
|---|---|---|---|
| `open:false` desktop 1600 | no | built, 17 rooms | pass |
| `open:true` desktop 1600 | yes | floor mounts 5 cabinets, 0 sheets | pass |
| never answers, 5.6 s | no | built, no exception | pass |
| `open:false` phone 390 lite | no | built | pass |

Zero exceptions in any log. `dotnet build`: 0 errors (322 standing warnings).

Shot of the closed alley: `C:\Users\PC\AppData\Local\Temp\claude\bk-g1-shots\g1-closed-d1600.png`

## Merge order

After #820 (integration). Pairs with server #139, which must deploy before the flag means anything; until then every client sees no door, which is the intended default.

Operator steps to open (env var names only): set `CASINO_TESTERS` to a comma list of account ids for the tester wave, or `CASINO_OPEN=on` for everyone. Server #139 documents both in `proxy/docs/env-vars.md`.

https://claude.ai/code/session_018oDQzrDxMaHFUuXnnY8c8W

### DIFF
--- a/ConditioningControlPanel/Resources/web/arcademy/backroom/cage.js
+++ b/ConditioningControlPanel/Resources/web/arcademy/backroom/cage.js
@@ -196,6 +196,7 @@ export function createCage(opts) {
   function refusal(r) {
     const body = (r && r.body) || {};
     const reason = String(body.reason || body.code || '');
+    if (reason === 'casino_closed') return t('bk_closed');   // a 403 too, but the house's, not the account's
     if (r && r.status === 403) return t('bk_cage_locked');
     switch (reason) {
       case 'insufficient_sparkle': return t('bk_cage_poor');
--- a/ConditioningControlPanel/Resources/web/arcademy/backroom/lex.js
+++ b/ConditioningControlPanel/Resources/web/arcademy/backroom/lex.js
@@ -52,6 +52,7 @@ export const BK_LEX = Object.freeze({
      the two rooms must not disagree about what that means, so this row is a
      copy of it rather than a new sentence. */
   bk_cage_locked:    'The bank does not serve this account yet. Nothing was charged.',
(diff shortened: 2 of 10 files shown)


CodeBambi:

PR #838, opened 2026-09-05
fix(logging): smoke-test fixes (footer on exit, literal message rendering, video-diag redaction)

CodeBambi:
Four defects found by a runtime smoke test of the logging redesign, plus the tests that pin them.

Branched from `feat/log-session-files` (the tip of the pipeline stack #825 -> #829 -> #831 -> #835 -> #836), with `main` merged in as the first commit. `main` carries the already-merged call-site PRs of this same programme, including `Services/Diag.cs`, which the footer needs; the merge is clean and there is nothing hand-resolved in it. My own diff is the three commits after it.

## D1 - the footer was never written

`LogPipeline.HookFooter` registered `WriteSessionFooter` on `AppDomain.ProcessExit` only, and `App.OnExit` ends in `TerminateProcess`, which skips ProcessExit handlers and finalizers by design (that is what stopped WPF's DirectWriteForwarder crashing on close, so it stays). `WriteSessionFooter` had no other caller. Three clean exits in the smoke test produced zero `== end:` lines.

`App.OnExit` now calls it directly, just before `Log.CloseAndFlush`. The `AppDomain.UnhandledException` handler calls it too when `args.IsTerminating` - ProcessExit does not run for those either, so the log of the run that crashed was the one with no end line. The ProcessExit hook stays for the exits `OnExit` never sees (taskbar kill, restart-for-update). The guard is now an `Interlocked.Exchange`, so whichever path arrives first is the one that writes.

Deliberately NOT wired into the dispatcher and unobserved-task handlers: both mark the exception handled and the app keeps running, where writing the footer would close the sinks under a live session.

## D2 - every string parameter was quoted

`CcpLineFormatter` used `LogEvent.RenderMessage`, which renders string scalars with quotes. The old sink was an output template carrying `{Message:lj}`, so nobody had ever seen them. On disk:

```
Application starting v"6.9.1"
lang='"en"' mod="none"
UI unresponsive for 15s""
```

and, worse than cosmetic, `"[BARK]" blocked ...` landed in category `App`, because `CategoryEnricher` reads the tag off the front of the message and a quoted tag is not a tag.

New `Services/Logging/LiteralMessage.cs` renders the template with strings written raw and everything else (numbers, enums, sequences, structures, formats, alignment) left to Serilog. Both the formatter and the enricher use it, so the column and the line can never disagree. Redaction is unaffected: `RedactingEnricher` rewrites property values before either caller renders, and the formatter still runs the redactor over the finished line.

## D3 - the footer omitted the swallowed count

`== end:` now carries `swallowed={n}` from `Diag.SwallowCount`, plus the busiest sites as one line each while that list stays at five or fewer. Those exceptions stop being logged once a site passes its per-site budget, so the footer is the only place a run ever says there were four hundred of them.

The count reaches the pipeline through a `SwallowProvider` seam rather than a direct reach into app-wide state, so the footer tests can supply their own numbers.

## D4 - video-diag.log wrote absolute user paths

`VideoDiag` bypasses Serilog entirely - it owns its thread and its file so the trace survives a wedged UI - so no pipeline redaction reached it, and `UiHangWatchdog` hands it raw paths at two call sites. Fixed at the sink so the next caller cannot reintroduce it.

Redaction runs on the writer thread, not in `Log()`: `Log()` is called from the UI thread and from LibVLC's native callback threads while the UI may already be wedged, and the redactor's assets-root rule does a `Directory.Exists` that can block on an unplugged drive. The redactor is otherwise regex-only, with no locks, no Serilog and no shared state beyond its cached roots.

## Not in scope

D5 (`CCP_USERDATA_DIR` not honoured by the Roaming `achievements.json` or by WebView2's `browser_data_video`) is left alone.

## Verification

`dotnet test` on the whole suite: 5280 passed, 0 failed. Logging classes alone: 280 passed.

Runtime, on `main` + this branch with `CCP_USERDATA_DIR` pointed at a sandbox: see the verification notes on the task. Footer present with `swallowed=`, `Application starting v6.9.1` unquoted, no stray `""`, `%DATA%`/`%APP%`/`~` intact, and no `C:\Users` in either the session file or video-diag.log.

Co-Authored-By: Claude Fable 5.1 <noreply@anthropic.com>

### DIFF
--- a/ConditioningControlPanel/App.xaml.cs
+++ b/ConditioningControlPanel/App.xaml.cs
@@ -1699,6 +1699,18 @@ namespace ConditioningControlPanel
             {
                 var ex = args.ExceptionObject as Exception;
                 LogCrashDetails("DOMAIN", ex);
+
+                // Last chance to close the session file: the runtime is about to tear the process
+                // down and ProcessExit does not run for an unhandled exception, so without this
+                // the log of the run that CRASHED is the one with no end line. Guarded on
+                // IsTerminating, and deliberately NOT done in the dispatcher or task handlers -
+                // both of those mark the exception handled and the app keeps running, where
+                // writing the footer would close the sinks under a live session.
+                if (args.IsTerminating)
+                {
+                    try { Services.Logging.LogPipeline.WriteSessionFooter(); }
+                    catch { /* swallow: nothing useful is left to report it to */ }
+                }
             };
             TaskScheduler.UnobservedTaskException += (s, args) =>
             {
@@ -4953,7 +4965,16 @@ Application State:
             SecureAuthTokenStore.ClearMemoryCache();
             SecureApiKeyStore.ClearMemoryCache();
 
-            // Close and flush the logger
+            // Close the session file and write its "== end:" line. This has to happen HERE rather
+            // than only from the pipeline's ProcessExit hook, because OnExit finishes with
+            // TerminateProcess (see the comment at the bottom of this method), which skips
+            // ProcessExit handlers entirely - so every ordinary close was ending with no footer at
+            // all. Idempotent: the ProcessExit hook and the unhandled-exception path can still call
+            // it, and whichever arrives first is the one that writes.
+            Services.Logging.LogPipeline.WriteSessionFooter();
+
+            // Close and flush the logger (WriteSessionFooter has already done this; harmless and
+            // kept so the shutdown reads the same whether or not the pipeline was ever built).
             Log.CloseAndFlush();
 
             // Dispose show-window signal
--- a/ConditioningControlPanel/Services/Logging/CategoryEnricher.cs
+++ b/ConditioningControlPanel/Services/Logging/CategoryEnricher.cs
@@ -18,6 +18,11 @@ namespace ConditioningControlPanel.Services.Logging
     /// literal from the call site, so the same string instance comes back every time. The cache is
     /// bounded and cleared wholesale when it fills, which cannot happen with a fixed set of call
     /// sites but can if someone builds templates by concatenation.</para>
+    ///
+    /// <para>The one template shape the text alone cannot answer is a tag passed as a PARAMETER
+    /// (<c>"{Tag} blocked"</c>). Those are rendered through <see cref="LiteralMessage"/> - the same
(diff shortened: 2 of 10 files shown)


CodeBambi:
Runtime verification, on the integration `origin/main` + this branch (local tag `int-log-0905b`, `39f33e6bc`), launched with `CCP_USERDATA_DIR` pointed at a sandbox and exited via the Exit button.

**Before** (pre-fix run, `session-20260905-083139.log`): 40 of 99 lines carried a double quote, and the file simply ended:

```
08:31:39.222 INF [App          ] Application starting v"6.9.1" | workingSet 116MB
08:31:39.723 INF [App          ] Loaded 5434 strings for language '"en"'
08:34:10.179 INF [Autonomy     ] All active pulses cancelled     <- last line, no footer
```

**After** (`session-20260905-090827.log`): 0 double quotes in 98 lines, 0 `C:\Users`, `%DATA%` x7 and `%APP%` x7 intact, and the file ends where it should:

```
09:08:27.313 INF [App          ] Application starting v6.9.1 | workingSet 115MB
09:08:27.960 INF [App          ] Loaded 5434 strings for language 'en'
== end: uptime 0:01:00 warn=0 err=0 suppressed=0 swallowed=0 ==
```

**Hang run** (`--test-ui-hang 20`, recovered normally). New `video-diag.log` lines:

```
09:10:48.818 [t8] WATCHDOG: hang report written: %DATA%\logs\hang_20260905_091048.txt
09:10:52.258 [t8] WATCHDOG: hang dump written (external): %DATA%\logs\hang_20260905_091048.dmp
```

against the same two lines from the pre-fix run, which carried the full `C:\Users\PC\AppData\Local\Temp\...` path. The auto-dumped `diag-20260905-091048-hang.log` has 0 quotes across 186 lines, where its pre-fix counterpart had 107.

One caveat on `swallowed=`: it reads 0 because nothing in the app calls `Diag.Swallowed` yet - `feat/log-swallow-catches-1/2/3` are still open. The plumbing is covered by `SessionLogTests.Footer_StatesWhatTheRunCost` (91) and `Footer_ListsTheBusiestSwallowSites_WhileTheListIsShort`; the number will start moving on its own once those land.

Tests: 5280 passed / 0 failed on the full suite (280 in the logging classes alone).

PR #941, opened 2026-09-07
feat(race): the popped word flashes on the caption band

CodeBambi:
W3 of the word bubbles wave, on top of W1 (#938) and W2 (#939).

The words are on the road now, so the caption band stops being where the script is READ
and becomes where a pop is ANSWERED. The typewriter goes one release behind `?cap=type`.

## what lands

**The flash.** `race/captions.js` grows `showWord(text, { ghost, accent, ink })`. The word
lands on the band whole, no typing, holds `FLASH_HOLD_MS` 500, lets go over `FLASH_FADE_MS`
300. A pop inside `FLASH_JOIN_MS` 300 of the last one JOINS its line, so a line of word
bubbles taken clean reads back as the sentence that was said; a pop after the line began to
go starts a fresh one. Two lines and no more: a word that spills the line into a third row
starts the line again rather than shrinking to fit, because the sentence has already been
read and this is the answer to a pop, not the script. Accent words (the `ACCENT_WORDS` list
`cues.js` already stamps on the spawn) are 1.3x, wear the ink of the set the road says that
second belongs to, and glow for 160 ms. Reduced motion fades in place and never zooms.

**The ghost.** A word bubble the kart drove past still writes itself on the band, grey and
at `GHOST_ALPHA` 0.35, so the script is never hostage to the steering. `?words=unread` turns
the ghosts off, for the owner to feel both.

**The typewriter** is untouched behind `?cap=type`. A track with no words file never had a
caption and still does not.

**The plate** is exactly what it was, on all thirteen themes, fired by `sync.js` at
`event.t` as today. What is new is that it CLEARS the band under it: one thing at a time
on the glass.

**The wiring** lives in `run.js`, not in `bubbles.js`. The pop and miss callbacks carry an
`eventId` but not the word, so `run.js` keeps the word each word bubble wears by the event
that spawned it (a ring of `WORD_MEM` 256, cleared with the run and on a track swap) and
spends it on the pop or on the pass. `race/bubbles.js`, `race/wordTags.js` and
`race/wordFace.js` are not touched by this PR: they belong to the parallel word-face task.

## what the brief asked for and the code already had

The last item of W3 was "on worded tracks the random effect dressing appears only where no
word bubble is due for over 4 s". That is already true, and more strictly than asked:
`bubbles.js` `sparse` (set from `TR.lyrics`) makes `seedChunk` place ONE golden per chunk
and nothing else on a transcript road, `spawnAhead` and the rain bursts stand down whenever
a track is loaded, and `cues.js` halves a peak's rain and draws it from a fixed room kind
rather than `rollKind`. There is no random effect dressing left on a worded road to gate,
so nothing was built for it.

## smokes

`race/smoke/captions-check.mjs` boots the same fixture twice: `?cap=type` for the typewriter
sections it always had, and the default build for the flash. The flash section holds the
whole rule (the word on the band within 50 ms of a pop, joined inside 0.3 s, a fresh line
after the fade, the ghost at 0.35, the accent bigger, no typewriter, two lines at most under
twelve long pops, and never a pixel on the score plate, the toast rail or the plate's rest
spot at 390x844), watches the band at the exact frame a plate lands, then lets the real road
roll through its first spoken words and counts what reached the band off real pops and passes
(16 on this run). It still never prints a word of a transcript: everything is counted.

Green on this head: `captions-check` (all good, 0 console errors), plus `chart-check`,
`words-check`, `words-rate-check`, `levels-check`, `cloud-chart-check`, `lyrics-road-check`,
`track-run-check`, `rows-check`, `sync-check`, `fov-check`, `pickups-check`,
`spiral-pool-check`, `label-px-check`, `wall-dom-check`, `poses-smoke`, `gltf-smoke`,
`audio-check`, `ost-resident-check`, `cloud-check`, `remote-feed-check`, and `pace-check` /
`touch-check` run standalone.

346 changed lines.

https://claude.ai/code/session_01EJ66hi5fRM1Dcjo38aSCa2

### DIFF
--- a/ConditioningControlPanel/Resources/web/dtrh/race/captions.js
+++ b/ConditioningControlPanel/Resources/web/dtrh/race/captions.js
@@ -11,15 +11,23 @@
  * Two layers, one file, because they are one idea: the road says the word, the
  * glass says the word, and the plate says it loudly.
  *
- *   THE CAPTION. `chart.words` (the caption track L2 puts on the road) is cut into
- *   phrases, one on screen at a time, low on the glass. Every word of the phrase is
- *   in the DOM from the moment the phrase opens but only INKED when the clock
- *   reaches its own timestamp, so the line types itself at the speed the voice
- *   actually spoke it and never reflows while it does. That is the typewriter: the
- *   voice is the timer, not a timer.
+ *   THE FLASH (`mode: 'flash'`, what a worded road runs). The words are on the ROAD
+ *   now, one to a bubble (race/wordBubbles.js), so the band is no longer where the
+ *   script is read: it is where a POP is answered. The word the kart just took hits
+ *   the band whole, holds FLASH_HOLD_MS, fades over FLASH_FADE_MS, and a pop inside
+ *   FLASH_JOIN_MS of the last one joins the same line, so a line driven clean reads
+ *   back as the sentence that was said. A word the kart drove past writes itself as
+ *   a GHOST, grey and faint: the script is never hostage to the steering.
+ *
+ *   THE CAPTION (`mode: 'type'`, behind `?cap=type`). The old typewriter: `chart.words`
+ *   cut into phrases, one on screen at a time, every word in the DOM from the moment
+ *   the phrase opens but only INKED when the clock reaches its own timestamp, so the
+ *   line types itself at the speed the voice actually spoke it and never reflows
+ *   while it does. The voice is the timer, not a timer.
  *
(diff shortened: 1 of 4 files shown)


CodeBambi:
Rebased onto main after #942 landed (one conflict, in the run.js public object: kept `wordFaces()`, dropped the deleted `wordTags()`, kept `debugWord()` and `capMode()`). All race smokes green on the new head efb191667, including the new face-check.

#943 (W4: the plate flies on the pop, the ladder counts lines) is stacked on this branch.

PR #1029, opened 2026-09-10
feat(neutral): a fresh install with no mod is neutral by default (Wave 1)

CodeBambi:
## Why

A long-time tier 2 supporter said the core app increasingly assumes a cis male user and that gendered content lands in unavoidable main-app space, and that Circe's Lock (the earlier answer) added a second gender rather than a neutral floor. A seven-slice audit confirmed it: the neutral layer already exists (`VocabTokens` defaults, the CCP Default mod identity), but about twelve independent fallback arms resolve an unmodded install to Bambi or Sissy instead.

This is Wave 1 of 4: flip those defaults. No renames, no persisted id changes, no new content beyond the two intake assets.

## What changes

- **Companion / AI**: new "CCP Default" persona preset (first in the picker). Persisted default persona id is neutral. The shared prompt builder no longer describes the vanilla user as a bimbo. The neutral baseline no longer inherits Bambi's recorded voice-over.
- **Voice commands**: a "neutral" pack, 43 lines including the panic confirmation, with a parity test against the sissy pack so it can never go silent. Drone keeps the pack it already used.
- **Possession / emergency exit**: vanilla rewrite and toast pools. Quit-flow honorific is the `VocabTokens` pet name on both C# and web; "Confirm you are NOT a good girl" can no longer render on vanilla.
- **Fresh-install pools**: subliminals, lock cards and triggers copy from the neutral CCP Default pools that already existed in `BuiltInMods`. Marquee default is "CONDITION DAILY" with a one-shot migration that only rewrites the exact old string; customised marquees are untouched.
- **Intake**: real "default" niche with `pass_card_default.png` and a 17-prompt `banks/default.json`. Every bambi fallback in the engine, shim and host service points at it.
- **Quiz / race**: neutral trick question and praise audio on vanilla; race structure words split into a neutral core plus themed extras; race end card no longer says "good girl".

## What does not change

Bambi Sleep, Sissy Hypno, Circe's Lock and Drone keep their own packs, pools, niches and presets. Existing users keep their saved pools (fresh-install defaults only apply on first run).

## Known, on purpose

- A few lines were never mod-gated, so themed users see the neutral text too: Deeper seeded praise ("that's it, sweetie"), the gaze minigame flash, two Autonomy praise lines, the DTRH freeze card and race end tags. The race layer gets a mod signal in Wave 3.
- Drone Mode and creator mods used to borrow the Bambi Possession pools and now get the vanilla one.
- Unmodded users have zero idle voice lines until Wave 4 ships a neutral audio pack.
- `pass_card_default.png` is a recolour of the Bambi card, not real art.
- Stale docs to fix in a follow-up: `EMERGENCY_EXIT.md:32,60`, `race/CHART.md:61,195`.

## Verification

- `dotnet build` 0 errors
- 251 tests pass, incl. new `VoiceCommandNeutralPackTests`
- race smokes green except the pre-existing `kinds-check "with a picture in it"`
- headless intake run on the default bank, 89 beats, clean

https://claude.ai/code/session_01LxWPhmMRDRwcHiHaon3MaE


CodeBambi:
Merged current `origin/main` into `feat/neutral-core-wave1` (`ffcca381e`). Ten files conflicted.

Context for anyone reading the diff: main already has this wave. It went onto the release line as cherry-pick `22932f7b0`, then main developed it further in `c72ddd8f0` ("the four wave 1 lines a mod user lost come back through the mod") and in the 6.9.4 copy pass. So on every conflict both sides want the same outcome and main carries the more complete mechanism: neutral wording kept as the default, but resolved through the active mod so a themed install does not lose its voice.

What main's cherry-pick did **not** take were the five race files, since the race is not on the release line. `Services/Race/TrackLexicon.cs`, `Resources/web/dtrh/race/chart.js`, `race/cues.js` and the two smoke checks merged clean and the wave's neutral-core / themed-extras split is still in the tree.

| File | Resolution |
| --- | --- |
| `Lab/GazeMinigame/GazeMinigameWindow.xaml.cs` | main. `App.Mods?.GetGazeCorrectMessage() ?? "GOOD"` replaces the wave's flat `"GOOD"`. Same neutral default, plus a mod hook. |
| `MainWindow/MainWindow.Marquee.cs` | main. Both call sites use `ResolveDefaultMarqueeMessage()` (active mod banner, else `AppSettings.DefaultMarqueeMessage`) and the migration is passed the active mod id. Two gates for one string, collapsed to main's. |
| `Models/AppSettings.cs` | main. `MigrateMarqueeMessage(string? activeModId)` replaces the wave's no-arg version: rewrites only on CCP Default, latches the one-shot flag either way, defers if the mod layer is not up yet. The wave's `DefaultMarqueeMessage`, `LegacyGenderedMarqueeMessage` and the migrated flag are untouched. |
| `Resources/web/dtrh/game/payloadFx.js` | main. Doc comment only; the neutral `FREEZE` default and the `WORDS` pool are identical on both sides. Main's wording says "the caller's own phrase" because the tube uses the card too. |
| `Services/Deeper/SpeakPromptSession.cs` | main. `VocabTokens.PetName` instead of `VanillaPetName` for the seeded praise. |
| `Services/Quiz/IntakeNiche.cs` | main. `Current()` delegates to the pure `Resolve(modId, tags, sissyContentMode)`, which also restores the `BambiSleep` branch the wave's fallback change had left unreachable. The neutral `"default"` fallback is unchanged. |
| `Services/Quiz/QuizSessionGenerator.cs` | main. The wave's neutral obedience / submission / catch-all pools are word for word the same, now wrapped in `ThemedOrNeutral(category, neutral)` so a mod's manifest phrases win. |
| `Services/Video/VideoService.cs` | main. The troll replay line goes through `GetAttentionCheckTrollMessage`, with the wave's neutral text as the default. |
| `Views/Deeper/DeeperEditorWindow.Unified.cs` | main. `VocabTokens.PetName` for the seeded praise. |
| `Views/Deeper/DeeperEditorWindow.xaml.cs` | main. Same swap, and it is the side that now builds the ToolTip example at runtime. |
| `Views/Deeper/DeeperEditorWindow.xaml` | main. The static `e.g. '...'` example comes out of the ToolTip because the code-behind sets it with the live pet name. The other two-gates case, collapsed to the code-behind. |

No language JSON conflicted. Every line the wave added was checked against the merged tree: the only ones absent are exactly the literals main rewrote.

`dotnet build`: 0 errors. `dotnet test Tests/ConditioningControlPanel.Tests`: 6153 passed, 0 failed.

PR #1066, opened 2026-09-11
fix(copy): pull the hardcoded english out of the feature panels

CodeBambi:
Copy pass, surface A, part 2 of 2. Stacked on #1065, so review that one first.

Part 1 rewrote strings that were already loc keys. This one deals with the other half of the problem: thirty user-facing strings in the feature control panels were plain English sitting in the XAML, so every non-English user read them in English no matter what language they picked. They are loc keys now, with a real translation in all nine files, and the copy got the same voice pass on the way through.

Three of them carried em-dashes and two more were shouting `Turn ON` / `Turn OFF` at people.

## Em-dashes removed from visible UI

Beyond the thirty keys, three more em-dashes were sitting in hardcoded chrome in these same files and are now plain hyphens: the `THE LOOM - weave your own spiral` and `CORNER GIFs - pin a GIF to a screen corner` buttons in `SpiralFeatureControl.xaml`, and the "no panic key set" glyph `🔑 -` in `SystemFeatureControl.xaml`.

## Fact checks

No string in this PR states a number or a behaviour that changed. The two longest technical tooltips (GPU decode, solid mode) were re-read against the code they describe and their claims were left intact, only re-worded.

## Before / after (en)

| Key | Before (hardcoded) | After |
|---|---|---|
| `tooltip_flash_solid_mode` | Draws flashes on one shared overlay instead of a new window per image, so fullscreen games stop reacting to the window churn. Flashes become click-through (no mouse pop or hydra clicks) [em-dash] gaze pop and stare-to-keep still work. Applies to the next flash. | Draws every flash on one shared overlay instead of a new window each time, so fullscreen games stop reacting to the churn. Flashes go click-through, so no mouse pop and no hydra clicks, but gaze pop and stare to keep still work. Takes effect on the next flash. |
| `setting_flash_solid_mode` | Solid mode (game friendly) | Solid mode (game friendly) |
| `setting_stare_to_pop` | Stare to pop | Stare to pop |
| `tooltip_stare_to_pop` | Look at a flash for ~1s to dismiss it (same effect as clicking, including hydra). Covers flashes only. Bubbles have their own Stare to pop switch on the Bubble Pop page. | Look at a flash for about a second and it goes, same as clicking it, hydra included. Flashes only. Bubbles have their own switch on the Bubble Pop page. |
| `setting_stare_to_keep` | Stare to keep | Stare to keep |
| `tooltip_stare_to_keep` | Keep a flash alive by looking at it. Stops boosting when you look away. | Looking at a flash keeps it around. Look away and it starts running out again. |
| `setting_linger_boost` | Linger boost | Linger boost |
| `tooltip_mind_wipe_custom_audio` | Pick your own mind-wipe sound. Keep it short [em-dash] about 2 seconds works best. Leave it on Default to use the built-in clips. | Pick your own mind wipe sound. Short works best, about two seconds. Leave it on Default for the built-in clips. |
| `label_mind_wipe_audio_default` | Default (built-in clips) | Default (built-in clips) |
| `btn_select_audio` | Select audio… | Select audio… |
| `tooltip_spiral_library` | Pick the spiral used everywhere. Drop your own .gif / image files in the folder, then Refresh. | Pick the spiral the whole app uses. Drop your own GIFs or images in the folder, then hit Refresh. |
| `btn_spiral_open_folder` | Open folder | Open folder |
| `label_no_spirals_yet` | No spirals in your folder yet [em-dash] only the built-in spiral is available. Click "Open folder" to add your own. | Nothing in your spirals folder yet, so you get the built-in one. Open folder to add your own. |
| `setting_force_video_gpu_decode` | Force video GPU decode | Force video GPU decode |
| `tooltip_force_video_gpu_decode` | Play mandatory videos on the GPU. Off by default (software decode) to dodge a white-screen bug on some PCs. Turn ON if a video only plays on one of your screens, keeps swapping screens, or freezes when it starts. | Decode mandatory videos on the GPU. It is off by default because some PCs go white-screen with it on. Worth turning on if a video only plays on one screen, keeps hopping between screens, or freezes as it starts. |
| `setting_blurred_video_background` | Blurred video background | Blurred video background |
| `tooltip_blurred_video_background` | Fill the empty bars around a video that doesn't fit the screen (e.g. a vertical clip on a widescreen monitor) with a blurred, zoomed copy of the video, like TikTok. Turn OFF for plain black bars. | When a clip does not fit the screen, a vertical one on a widescreen monitor say, the bars get filled with a blurred copy of the video. Off gives you plain black bars. |
| `setting_subliminal_solid_mode` | Solid mode (game friendly) | Solid mode (game friendly) |
| `tooltip_subliminal_solid_mode` | Draws subliminal text on one shared overlay instead of a dedicated window per screen - lighter on the renderer and friendlier to fullscreen games. Ignored while 'steal focus' is on. Applies to the next flash. | Draws subliminal text on one shared overlay instead of a window per screen. Lighter to render and kinder to fullscreen games. Ignored while steal focus is on. Takes effect on the next one. |
| `tooltip_subliminal_font_any_installed` | Any font installed on Windows. Applies to the next subliminal. | Any font installed on Windows. Applies to the next subliminal. |
| `tooltip_bouncing_font_any_installed` | Any font installed on Windows. Applies live - no restart. | Any font installed on Windows. Applies live, no restart. |
| `setting_bubble_solid_mode` | Solid mode (many bubbles) | Solid mode (many bubbles) |
| `tooltip_bubble_solid_mode` | Uses the Rabbit Hole's optimized renderer so the field stays smooth with lots of bubbles. Lifts the on-screen cap. Turn off to use the classic one-window-per-bubble mode. Takes effect when bubbles restart. | Borrows the Rabbit Hole's renderer so a crowded field stays smooth, and lifts the on-screen cap. Off goes back to one window per bubble. Takes effect when bubbles restart. |
| `setting_trigger_bubbles` | Trigger Bubbles | Trigger Bubbles |
| `tooltip_trigger_bubbles` | Some bubbles fire an effect when popped (flash, video, overlays…) | Some bubbles set something off when you pop them: a flash, a video, an overlay. |
| `label_trigger_bubbles_egg_hint` | careful [em-dash] {persona} loves these… | careful, {0} loves these… |
| `setting_duration_filter` | Duration filter | Duration filter |
| `tooltip_duration_filter` | Duration filter: min length → max length. Filter videos by length. 0 = no limit. | Only play videos between these two lengths. 0 means no limit. |
| `setting_stare_to_click` | Stare to click | Stare to click |
| `tooltip_stare_to_click` | Look at a target for ~1s instead of clicking. Same hit, same reward. | Look at a target for about a second instead of clicking it. Same hit, same reward. |

Where before and after match, the string was already fine and the only change is that it now reaches a translator. Titles and control names were left alone per the brief.

## The Trigger Bubbles hint

`BubblePopFeatureControl.xaml.cs:160` built the string in C#, persona name and all:

```csharp
TxtTriggerEggHint.Text = $"careful — {persona} loves these…";
```

It is now `Localization.Loc.GetF("label_trigger_bubbles_egg_hint", persona)`, so the wrapper translates and the persona name still comes from the active mod. The em-dash is gone with it.

## Reused rather than added

The `Gaze interaction` heading in `FlashFeatureControl.xaml` now points at the existing `label_gaze_interaction` key, whose value already matched exactly, rather than adding a thirty-first key.

## Verification

- `dotnet build` 0 errors.
- `dotnet test` 5835 passed, 0 failed.
- All 30 keys are new: the script refuses to write a key that already exists in a file, and a post-check confirms each appears exactly once in all nine files and none of them collide with the duplicates the repo already carries.
- Every language file still parses strictly, with `\n` escapes and LF endings untouched.
- 380 changed lines, under the 600-line cap.

## Still hardcoded, left for a follow-up

These are user-facing English in the same files but fell outside the 30-key budget: the Bubble Pop `Trigger chance` / `Effect types` labels and their `Subliminal` / `Pink Filter` checkboxes, the bubble speed tooltip, `LockCardFeatureControl`'s spoken-phrase hint, and `FeaturePopupWindow`'s `Close (Esc)`.

https://claude.ai/code/session_013mR8cAipk1ivr3Em5fJjcx


CodeBambi:
Merged into release/6.9.4 via its branch merge; GitHub keeps a stacked PR open once its base branch is merged.

PR #1114, opened 2026-09-13
feat(backroom): C4 media feed (BackRoomMedia)

CodeBambi:
Lane C4 of the Back Room, CONTRACT.md section 5. Implements `IBackRoomMedia`.

**What it deals**
- `g0..g3`: local animated files from the DISK half of the flash pool (new `FlashService.SnapshotLocalImagePaths`, same refresh and deselection prune as `GetChaosImagePaths`). Remote urls dropped (`IsRemotePath`), pack temp copies dropped (no `ccp.assets` url behind them), anything outside `EffectiveAssetsPath` dropped. Deduped by FULL path (case-insensitive), sorted, then a seeded partial Fisher-Yates. `.gif` by magic, `.webp` only with the VP8X animation flag; width and height from the same 30-byte read, bounded at 48 reads per deal. Urls encoded like `ArcademyHostService.ToAssetsUrl`, animated webp carries the `#.gif` hint.
- Shortfall: `https://ccp.game/backroom/stations/slot/fallback/gifN.webp`, four 180x180 loops cut from the existing dtrh spiral gifs (no generation).
- `s0..s3`: enabled `SubliminalPool` words (the live per-mode / per-mod pool), own seeded stream, then presets Drop, Relax, Let Go, Sink as lexicon keys `br_word_drop`, `br_word_relax`, `br_word_let_go`, `br_word_sink` with neutral fallbacks.
- Logging: one Information line with counts only; failures log the exception type, never its message.

**Tests** (`BackRoomMediaTests`, 8): privacy (deal JSON and captured Serilog output carry no root, drive path or file name; a throwing listing with a path in its message does not leak), full-path dedupe, shortfall fill, empty pools, remote-only pool, only-animated-under-root, seed stability across listing order.
`dotnet test --filter BackRoomMediaTests`: 8/8. Full suite: 6163/6163.

**For C1:** construct `new BackRoomMedia(lex)` passing the room lexicon resolver `(key, fallback) => text`; without one the neutral text is used. Call `Deal` off the UI thread (header reads).


CodeBambi:

PR #1130, opened 2026-09-13
Back Room R1c: swap the 2D picture room for the walkable 3D room

CodeBambi:
Stack R1 (3D room swap), 3 of 5. Base: `feat/br2-r1-room3d-b`.

**The swap.** `index.html` gets an importmap (`three` -> `/vendor/three/`, which C2's slot station also needs for its bare `three` imports). `room/scene.js` replaces the SVG room: renderer, lights, walk (WASD / arrows, Shift runs), drag look, E visits the nearest station, M toggles the room view (ceiling off, a list of stations). `room/main.js` boots it and routes Back / Escape through the open station, then the room view, then the room.

**Visiting.** E calls `scene.hold()` (pose saved, rAF stopped, WebGL context kept), then the loader opens the station or the dust-sheet card. Back closes it and `scene.release()` restores the exact position and facing. Measured: about 40 ms from Back to a drawing frame, against about 1.1 s to boot the room again, so the room keeps its context.

**Still.** Floor, bulbs, wall-picture turns, roulette hub and head sway freeze whenever init / settings say `reduced` or `intensity: calm` (the Motion button is locked then), or when the player picks Motion still.

**Registry.** `stations.json` rewritten for 3D from `placements.json`: counter, wheel, slot x3 (same station id `slot`, variants rose / violet / mint; violet and mint carry palettes), cards, roulette. Only the slot rows are live. No scratcher.

**Dropped from the 2D room:** the picture backdrop and SVG hotspots, and the click-to-walk student sprite (first person now). Not carried from the preview: the touch joystick and the local file picker. Kept: bridge, loader, soon card, Back / SP HUD, exits. The Play wall strip, palette row and EMI ring card (C1f, WPF side) still make sense and are untouched; `backroom_final.png` stays because the EMI ring card uses it.

This PR alone still carries the old smoke check; PR 4 replaces it.


CodeBambi:

PR #1131, opened 2026-09-13
Back Room R1d: ctx.variant, 3D smoke check and contract amendments

CodeBambi:
Stack R1 (3D room swap), 4 of 5. Base: `feat/br2-r1-room3d-c`.

- `room/loader.js`: `ctx.variant` = `{ id, name, palette }` for a variant row, `null` otherwise; soon and failed cards are titled with the row name. The 2D `geometry.js` is deleted.
- `smoke/room-check.mjs` rewritten for 3D, all passing: registry shape; every approach clear, nearest, and reachable from the door; glb node names on disk; boot with 7 fixtures, 4 screens, 161 bulbs and 212 draw calls; W walks and a drag turns; at every approach, walking forward stops outside the body; E on the violet slot opens the mock station with the violet palette, the room loop stops while it is open, and Back restores the exact pose; the rose slot opens the same station; the roulette gets the soon card and never posts station-open; M opens the room view; host close answers exit-done in about 50 ms; reduced motion and Calm freeze the floor and sway and a settings frame unfreezes it; a feed picture reaches the walls and the fallback entry is skipped; Escape leaves. It only ever stops the Chrome it spawned.
- A scratch run against the real C2 slot station (served read-only from `C:/wt-br2/c2-slot`) also passed: the station canvas mounted next to the held room, Back restored the pose, and the station canvas was removed.
- `CONTRACT.md`: sections 1, 6 and 7 amended, plus a dated 7.1 "3D room amendments".

**For C2:** `ctx.variant.palette` maps a material name to `rrggbb` (for example `candy_rose: ac83ed`). Recolour the close-up cabinet the way the room does and put `variant.name` on the marquee; ignoring it is safe. Also, C2's own Back chip sits under the room HUD's Back in the same corner (see `real-slot-station-open.png`).

Evidence: `C:/wt-br2/_evidence/r1-room3d/` (approach-*.png, slot-violet-station-open.png, soon-roulette-card.png, room-view.png, reduced-motion-still.png, wall-picture-from-feed.png, real-slot-*.png, room-perf.json).

**Test.** `node ConditioningControlPanel/Resources/web/backroom/smoke/room-check.mjs <evidenceDir>`


CodeBambi:

PR #1132, opened 2026-09-13
Back Room R1e: room lexicon for the 3D room

CodeBambi:
Stack R1 (3D room swap), 5 of 5. Base: `feat/br2-r1-room3d-d`.

Lexicon for the 3D room, with the English value in all nine language files (the repo convention):

- new: `br_station_slot_rose` / `_violet` / `_mint`, `br_station_roulette`, `br_visit`, `br_room_view`, `br_room_walk`, `br_motion_on`, `br_motion_still`, `br_loading`, `br_slot_jackpot`, `br_slot_status`, `br_wheel_status`, `br_ad_arcademy`, `br_ad_dtrh`, `br_ad_focus_gaze`
- values moved to the owner's names: `br_station_wheel` Daily Daze, `br_station_cards` Soft Hand, `br_station_counter` The Prize Parlour; `br_walk_hint` now describes WASD, drag and E
- `BackRoomHostService.LexKeys` sends the new keys in `init.lex`

The page has an English fallback for every key, so the stack below works without this PR. Not built locally (a string-array addition only).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01V3c2CJTNfpbeBYqwjT6ZAj

PR #1133, opened 2026-09-13
Back Room R1f: wall GIFs play in the room, and the room owns the only Back

CodeBambi:
Stack R1 (3D room swap), 6 of 6 (fix pass). Base: `feat/br2-r1-room3d-e` (#1132).

**Wall GIFs play.** `room/gif.js` decodes a dealt GIF in the page with WebCodecs `ImageDecoder` (shipped in Chromium since 94 and available in WebView2 on `https://ccp.game`, so nothing is vendored). It returns composited frames, which are drawn into a canvas texture. A page without it, or a file it refuses, falls back to the old still first frame.

Caps (`room/screens.js`):
- a picture advances only while a screen showing it is inside the camera frustum (never in the room view, never while a station holds the room)
- at most 12 frames a second, into a texture of at most 384 px on the long edge
- at most one new decode per rendered frame, one decode in flight per picture
- Still (reduced motion, Calm, Motion still) holds the first frame

**Perf, same headless test** (Chrome, 1280x720, vsync off, four dealt spiral GIFs, pose facing two screens; this run landed on the Intel iGPU):

| | frame median | p95 | decodes |
|---|---|---|---|
| before (branch e, still first frames) | 6.5 to 6.7 ms | 7.5 to 8.0 ms | 0 |
| after, playing | 7.4 ms | 9.3 ms | about 24 a second (two visible GIFs) |
| after, reduced motion | 7.2 to 7.8 ms | 8.3 to 8.8 ms | 0 |

A first try at 512 px and two decodes per frame gave p95 11.5 ms, so the caps were tightened. JS heap grows from about 26 MB to about 70 MB with four GIF decoders open. Evidence: `C:/wt-br2/_evidence/r1-room3d/perf-gif-before-after.json`, `wall-picture-from-feed.png`.

**One Back.** `room/loader.js` sets `ctx.hostBack: true`; the slot side is `feat/br2-c2-slot-d`.

**Smoke** (all passing): under reduced motion the GIF is dealt but holds its first frame, out of view it does not advance, in view it plays capped (10 frames in 1 s), and `ctx.hostBack` reaches the station. CONTRACT 7 documents `hostBack`; 7.1 gets the playing-GIF caps.

Asset rebuild: every blender-scripting source still matches the committed build (shell `36a56c0b`, slot `0dfac025`, wheel `52f7682b`, counter `93dba883`, card-table `a38e41f6`, roulette `7d0d4bfc`, ads unchanged), so there are no new assets.


CodeBambi:

PR #1128, opened 2026-09-13
Back Room R1a: room asset speed pass, optimized shell and fixture copies

CodeBambi:
Stack R1 (3D room swap), 1 of 5. Base: `feat/br2-c1-room-f` (#1124).

**What.** `Scripts/build-backroom-room-assets.mjs` copies the Back Room shell and the five fixture glbs out of `C:\Projects\blender-scripting` (read only; the script refuses to write there) and writes optimized copies into `Resources/web/backroom/room/assets/`:

1. strip textures the room replaces at runtime (wall screen ads, floor inlay)
2. dedup, lift static leaf meshes to their kept group (ceiling, dealers, EMI root), join per material, prune (attributes kept)
3. textures to WebP, max 1024 px; wall ads to 1280 px WebP
4. meshopt (reorder, quantize, EXT_meshopt_compression)
5. node check: every name the room reads keeps its count, or exit 1

Vendors `vendor/three/addons/libs/meshopt_decoder.module.js` from three r169 (upstream md5 `cd7a7b72371f13b93fa92a8d3176762f`, provenance header added). Pinned tools install once under `%LOCALAPPDATA%/ccp-tools`, not in the repo.

**Sizes (binary, not in the line count):**

| file | source | copy | mesh prims |
|---|---|---|---|
| shell.glb | 7.88 MB | 1.50 MB | 350 -> 26 |
| slot.glb | 0.60 MB | 0.19 MB | 136 -> 58 |
| wheel.glb | 1.76 MB | 0.67 MB | 195 -> 61 |
| counter.glb | 3.72 MB | 1.40 MB | 444 -> 33 |
| card-table.glb | 4.12 MB | 1.61 MB | 279 -> 29 |
| roulette.glb | 4.55 MB | 1.62 MB | 318 -> 71 |
| ads (3 webp) | 3.02 MB | 0.42 MB | |
| **total** | **25.65 MB** | **7.42 MB** | **1,722 -> 278** |

Triangles are unchanged (no simplification). `.glb` is not LFS-tracked; these are committed plain like `stations/slot/assets/slot.glb` and the dtrh / piecebypiece glbs (largest copy 1.62 MB). LFS is an open question for the owner, not changed here.

**Perf, whole room** (headless Chrome, 1280x720, entry pose, vsync off, same harness for both):

| | preview (blender-scripting) | client room (this stack) |
|---|---|---|
| draw calls | 1,268 (1,312 in its own check) | 212 |
[...]

PR #1127, opened 2026-09-13
feat(backroom): C2c slot station, media, fx, dev harness

CodeBambi:
Lane C2, part c of 3 (stacked on https://github.com/CodeBambi/Conditioning-Control-Panel---CSharp-WPF/pull/1126). The slot station module, media deal, fx wiring and the standalone harness. CONTRACT.md sections 2 (page side), 4, 5, 7.

## What
- `station.js`: `mount(ctx)` -> `{open, close, suspend, destroy}`. Back is live while loading, rising and spinning (no network wait). One outcome per press (lever, Spin, Space); 1/2/3 light a freeze column. Readouts: `shownSp`, jackpot from `table.jackpot`, last win / free spins / melt on the status line and the cabinet screens, odds panel from `table.lines`. Refusals: quiet card with Back. No state: "closed for a moment" and Back (no fallback table). Cursor flushed on close.
- `media.js`: `ctx.media()` on sit-down, `gif0-3` / `sub0-3` map to the dealt items, GIFs kept animating in a hidden holder and redrawn into the reel canvas every 100 ms, fallback art when an item is missing or fails to load. Only ccp.assets / ccp.game URLs (or the page's own origin for dev) are loaded.
- fx: every landed outcome sends `outcome.fx` in order with dealt KEYS (`g1`, `s0`), never urls; `melt {left}` whenever it changes. Nothing waits on `fx-ack`; fx are dropped while suspended.
- `dev.html` + `mock-server.js`: runs standalone. Serve `ConditioningControlPanel/Resources` as root, open `/web/backroom/stations/slot/dev.html` (`?sp=&melt=&floor=&latency=&reduced`).

## Evidence (`C:/wt-br2/_evidence/c2/`)
- `evidence-run.log` / `evidence.json`: 34/34 CDP checks at 1280x720 and 390x844 (driver `evidence.cjs`).
- Rise: `d1280_01_rise_250.png`, `_450`, `_900`, `d1280_02_seated.png`, `p390_02_seated.png`.
- 10-spin tape: `d1280_03_tape_spin1.png`, `d1280_04_tape_spin6.png`, `d1280_05_tape_done.png` (+ p390).
- Freeze mid-tape: `d1280_06_freeze_armed.png`, `d1280_07_freeze_spinning.png` (+ p390).
- Melt carry-over: `d1280_08_melt_landed.png`, `d1280_10_melt_reopened.png` (+ p390).
[...]

PR #1134, opened 2026-09-13
Back Room C2d: slot hides its Back in the room, variant colours, mock on table v5 and 3000 ms floors

CodeBambi:
C2 slot station, stacked on `feat/br2-c2-slot-c` (#1127). Pairs with the room's R1 stack (#1128 to #1132 plus the R1f fix pass).

**One Back in the room.** When the room passes `ctx.hostBack === true`, the station hides its own Back chip and the card's Back button (Escape still stands up) and moves its top readouts clear of the room's Back and SP chips. `dev.html` passes no flag and keeps its Back.

**Variant colours.** New pure `palette.js` recolours `candy_rose`, `candy_violet`, `candy_plum` and `wand_pink` from `ctx.variant.palette` with one clone per material name (the glb material is never edited). `scene.js` applies it right after load, and `station.js` puts `variant.name` on the marquee. `dev.html` takes `?variant=violet|mint`. Checked inside the 3D room with the real station: Candy Violet, Candy Mint and Candy Rose each reach play in their own colours with only the room's Back showing, and Back returns to the same pose (`C:/wt-br2/_evidence/r1-room3d/real-slot-{violet,mint,rose}-open-in-room.png`).

**Mock server on CONTRACT 10.** Table v5: `emi3` pays 400, odds "1 in 12,987", drawn at weight 77 per million on its own seeded stream (strip draws stay seed-stable, and a strip row never lands on EMI x3 by chance). Sampled at 1 in about 13,700 over 1.4 M outcomes. Floors are 3000 ms per outcome for tapes and freezes, and a freeze right after a freeze waits max(3000, its outcomes x 3000). `dev.html` now defaults to `floor=3000`.

**Tests.** `node --test tests/*.test.mjs` passes 20 of 20, including 2 new palette tests; `tests/nodes-check.mjs` is OK. (Passing the directory to `node --test` fails on Node 24, so use the glob.)


CodeBambi:

PR #1137, opened 2026-09-13
fix(backroom): room HUD buttons never keep Space and Enter

CodeBambi:
Back Room checkpoint 2 fix pass, bug 1 (HIGH, R1).

**Bug.** A clicked room Back kept keyboard focus (only the hud.js pills blurred). A later Space or Enter pressed it again: it closed the station, or the whole room.

**Fix** (room/main.js, room/hud.js)
- One delegated pointerup listener drops focus from every room HUD button (`.br-hud`, `#br-room-ui`), replacing hud.js's per-pill blur.
- While walking or with a station or card open, Space and Enter on a HUD button are eaten (preventDefault plus stopPropagation on keydown and keyup) and the focus is dropped, so the station's own keys reach it from the next press. The room view and a boot that never finished still take them.

**Test.** room-check.mjs section 2d2 does a real CDP mouse click on Back, then Space and Enter while walking, on a focused Back with a station open, on a focused Back while walking, and Enter on a focused Room view pill. On the old code it fails 4 checks (plus cascades): room-check-before-fix.log. With the fix all room checks pass: room-check-bug1.log.

Stacked on #1133 (R1 tip). Evidence: C:/wt-br2/_evidence/fix-cp2/


CodeBambi:

PR #1139, opened 2026-09-13
fix(backroom): room host maps ccp.content so the slot's pack sfx play

CodeBambi:
Back Room checkpoint 2 fix pass, bug 3 (MEDIUM, C1 host).

**Bug.** BackRoomHostService had no ccp.content mapping, so https://ccp.content never resolved, CCP_CONTENT_READY was never injected, and the slot's chimes and thud (the race's pack sfx through dtrh/shared/audioSrc.js) were silent.

**Fix.** Adds `ChaosWebViewHost.ContentMapping()` to the room's mappings, the same helper the Arcademy, DTRH and Caucus hosts use (it creates the folder first and maps it Allow).

**Home.** Stacked on the R1 fix stack because R1 sits on C1 and the lexicon fix above already edits this file; a PR on the C1 tip would conflict with R1's LexKeys edits.

**Not verified without the real app:** the WebView2 mapping itself and audible chimes. Build and the full test suite pass on the verify branch. Evidence: C:/wt-br2/_evidence/fix-cp2/


CodeBambi:

PR #1140, opened 2026-09-13
fix(backroom): the media deal runs off the UI thread

CodeBambi:
Back Room checkpoint 2 fix pass, bug 6 (optional LOW, C1/C4 seam).

**Bug.** BackRoomBridge.OnMediaRequest called `Media.Deal` inline from Handle (the UI thread), though BackRoomMedia reads file headers and says to call it off the UI thread.

**Fix.** `BackRoomBridge.Deps.OffUi` (null = inline, which the suite keeps). The host passes `Task.Run`; the deal runs there and the reply goes through the host's Post, which already marshals to the dispatcher. A deal that finishes after the window closed posts nothing.

**Test.** New BackRoomBridgeLifecycleTests.Media_DealsOffTheUiThread_AndPostsOnlyOnceTheWorkRuns. BackRoom tests 40/40, full suite 6515/6515 on the verify branch. Evidence: C:/wt-br2/_evidence/fix-cp2/


CodeBambi:

PR #1141, opened 2026-09-13
feat(backroom): room owns the SP chip rule, ctx.spReadout for stations

CodeBambi:
Back Room checkpoint 2 fix pass, bug 5, room side (LOW, R1 + F1).

**Bug.** On close the slot wrote ctx.sp() (server SP) into #br-sp-value while wins were still unplayed, so the chip dipped on Back and rose again on reopen. The room also never repainted the chip when a station-result moved state.sp.

**Fix** (room/main.js, room/loader.js)
- The chip always shows Law I shownSp = state.sp minus what the tape still owes, on every repaint (init, balance, station-result, close, reopen).
- New `ctx.spReadout = { set(value), owe(n), thud(), target() }`: `owe` takes a number or a reader `() => n`; `set` shows a BANK tick as is, `set(null)` returns to the rule; `thud` is the landing mini-thud (lit when reduced); `target` is the chip box for token flights.
- When a station closes, the loader freezes a reader to its last answer and drops any set value. A station-result repaints on the next frame, after the station has adopted the same reply.
- CONTRACT 7.1 amended: the SP chip rule and the HUD keys rule; section 7 ctx listing gains spReadout.

**Test.** room-check.mjs 2d3 drives the hook through the mock station (flight value, station-result, Back, balance frame, reopen); the ctx field check includes spReadout.

The station side is #1145 (stacked on the F1 tip). They need to land together: until then the slot still does its old id lookup. Evidence: C:/wt-br2/_evidence/fix-cp2/


CodeBambi:

PR #1118, opened 2026-09-13
feat(backroom): intensity row in Settings, lexicon keys, fx desk-run rig (C3d)

CodeBambi:
Lane C3d, stacked on https://github.com/CodeBambi/Conditioning-Control-Panel---CSharp-WPF/pull/1117. Stack: a -> b -> c -> **d**.

- Settings > Performance: "Back Room effects" row under Motion (Calm / Normal / Full), keys in all 9 language files.
- Dev trigger: `--backroom-fx-rig [outDir] [--fx ids] [--intensity Calm,Normal,Full] [--motion Full] [--all-toggles] [--exit]` fires each fx id at each intensity through the real dispatcher without the room, logs acks to `acks.jsonl`, grabs the screen per effect. Never writes settings. Tip: run it with `CCP_USERDATA_DIR` pointing at a sandbox so no real profile syncs.

Changed lines: 256. Full suite: 6447 passed.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01V3c2CJTNfpbeBYqwjT6ZAj

PR #1115, opened 2026-09-13
feat(backroom): fx recipe table and pure resolver (C3a)

CodeBambi:
Lane C3a (Back Room effects bridge, CONTRACT.md section 4). Stack: **a** -> b -> c -> d.

- `Services/BackRoom/BackRoomFxPlan.cs`: every fx id at Calm / Normal / Full, verbatim from the contract table ("+" = same start, "then" = after the previous piece).
- Pure resolver: Calm forced when MotionLevel != Full, per-primitive motion rules, feature toggles (Flash, Subliminal, Spiral, BrainDrain + melt toggle), symbol keys resolved only through the host's deal (unknown -> random dealt item). Every primitive that does not play is a reported skip (`calm`, `motion`, `toggle`, `unknown`).
- Tests: `BackRoomFxPlanTests` (intensity x MotionLevel x toggles over every id).

Changed lines: 599. Tests: `dotnet test --filter BackRoomFx` green.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01V3c2CJTNfpbeBYqwjT6ZAj

PR #1143, opened 2026-09-13
feat(backroom): room host fires the real fx dispatcher and deals the real media feed

CodeBambi:
Back Room checkpoint 2 fix pass, item 7: a home for verify wiring commit 004611342 (cherry-picked as is).

Replaces the C1 null objects in BackRoomHostService with `BackRoomFxServices.Shared` (C3) and `new BackRoomMedia(LexOrFallback)` (C4).

**Merge order: this needs the C3 fx stack (#1115-#1118) and the C4 media PR (#1114) merged first.** On this stack alone it does not compile, because those types live in those stacks. It builds and the full suite passes on the local verify branch, which merges everything. Evidence: C:/wt-br2/_evidence/fix-cp2/


CodeBambi:

PR #1144, opened 2026-09-13
fix(backroom): slot melt readout follows the tape cursor through a freeze

CodeBambi:
Back Room checkpoint 2 fix pass, bug 4 (LOW, C2 page).

**Bug.** The server settles a whole tape up front, so a freeze bought mid-tape carries `meltLeft` from the melt left at the END of the stored tape. tape.js land() copied it into the shown melt, so the readout jumped (for example "Melt: 3 spins at half" with no melt on screen yet), then jumped back when the tape resumed.

**Fix** (page only, server unchanged)
- tape.js land(): only a landing off the stored tape moves `melt`; a freeze and its re-spins (sealed from melt, 10.2) leave it where the tape's last landed spin put it.
- station.js land(): the landing beat (EMI rest pose, melted Brake, chime ladder) reads the tape's melt, not the freeze outcome's own `meltLeft`.

**Test.** New tape.test.mjs case: 9 blank rows then a melt, freeze mid-tape plus its re-spin: shown melt stays 0 and nothing is reported until the tape reaches the melt row. Fails on the old tape.js, passes now. Slot node tests 32/32.

**Verified** headless with the real slot: the status line never shows Melt through a mid-tape freeze (cp2-03-freeze-melt-stable.png); before the fix it showed "Melt: 3 spins at half" (before-fix/cp2-check-before.log). Evidence: C:/wt-br2/_evidence/fix-cp2/


CodeBambi:

PR #1145, opened 2026-09-13
fix(backroom): slot hands the room's SP chip what the tape owes, via ctx.spReadout

CodeBambi:
Back Room checkpoint 2 fix pass, bug 5, station side (LOW, F1).

**Fix** (station.js, station.css, tape.js)
- Uses the room's CONTRACT 7.1 `ctx.spReadout` instead of `document.getElementById('br-sp-value')` and its MutationObserver (both removed).
- Once its tape state is back it registers `owe(() => tape owed)`; while THE BANK flies it calls `set(value)`, then `set(null)`; the landing thud and token target come from `thud()` and `target()`.
- close() hands over the stored tape's unplayed pays as a number (`tapeOwed`, what a reopen will show; a freeze's own unplayed outcomes are not stored). A Back before the tape came back leaves the room's value alone.
- tape.snapshot() gains `owed` and `tapeOwed` (asserted in the mid-tape freeze test). `.slot-sp` hides on `data-host-sp`, so a room without the hook still shows the station's own chip; standalone dev.html is unchanged.

**Needs** the room side, #1141 (R1 fix stack), in the same build.

**Verified** headless with the real slot and 21 SP of wins unplayed: seated 39, after Back 39 at every 20 ms sample, 39 through the reopen load, 42 after the next +3 win (cp2-04/05/06 screenshots). Before the fix the chip read 60 after Back (before-fix/cp2-check-before.log). Evidence: C:/wt-br2/_evidence/fix-cp2/


CodeBambi:

PR #1156, opened 2026-09-14
docs(backroom): CONTRACT 10.13 Hypno v3 amendment

CodeBambi:
Docs only. Amends `ConditioningControlPanel/Resources/web/backroom/CONTRACT.md` with section 10.13 so the Hypno v3 build lanes (host H1, kit K1, stations G-wheel / G-cards / G-roulette, server S-cards / S-roulette) can work in parallel from one binding document. Spec: the owner-approved `hypno-spins-v3.html` mockup and its notes.

What it pins
- **10.13.A** `init.gates` `{flash, subliminal, spiral, brainDrain}` in `init` and every `settings` frame; `ctx.gates`, `ctx.onSettings`; what each station dresses plain per gate.
- **10.13.B** Which v3 effects are host overlays vs page effects. New ids `fx.wash`, `fx.gif_from`, `fx.loom_spiral`, `fx.haze`, the `fx.args` object, `fx-tunnel` (continuous level, 10 Hz, auto-release 1.5 s) and `fx-release`; Calm/Normal/Full recipes, gates, primitives, screens, coordinates. Spiral source: the player's own `loom_*.gif`, else bundled Loom-woven `shared/hypno/spirals/{screen,wake}.gif`; `spiral-full` moves to the same source.
- **10.13.C** `media-request.count` (1..13), the 13-GIF deal without fallback padding, value mapping `gifs[i % n]`, two-digit symbol keys.
- **10.13.D** Kit exports under `shared/hypno/`: `createLoomKit`, `LOOM_PRESETS`, `phaseForAngle`, `createDeck`, `DECK_VALUES`, `createMoments`, `MOMENTS`, `wheelSize`, `strengthK`; `room/gif-decode.js` split.
- **10.13.E** Soft Hand routes (state/deal/hit/stand/double/split) and Velvet Vortex routes (state/spin/cursor) on the simulated pure modules, with refusals, locks, receipts, floors (`CARDS_FLOOR_MS` 8000, `ROULETTE_FLOOR_MS` 6000), cap, `netSp`.
- **10.13.F** Moment table and station specs (file layouts, in-station effects, the Lighthouse beam on its own clock, handedness, 2D canvas tables for cards and roulette since no station glb contract exists).

Cross-reference edits in sections 1, 2.1, 2.2, 3 (Ops rows), 3.5 (wheel fx superseded), 4 (`spiral-full` source), 5 and 7 (ctx, third GL context).

[...]

PR #1164, opened 2026-09-14
feat(backroom): hypno v3 room wiring, gates, fx args, tunnel, media count (K1 1/6)

CodeBambi:
Lane K1, part 1 of 6 (stack: this, -b, -c, -d, -e, -f). Base: `feat/br2-hypno-contract` (#1156), CONTRACT 10.13.

**Room side of the hypno v3 wire**
- `room/loader.js` ctx gains:
  - `gates`: a live, frozen getter.
  - `onSettings(fn)`: returns an unsubscribe. Anything a station leaves subscribed is dropped when it closes.
  - `fx(fxId, symbols, args)`: its promise carries `.token` at once.
  - `fxRelease(token)` and `fxTunnel(level)`.
  - `media({count})`: count is sent only when it is an integer 1..13.
- `room/main.js` reads `init.gates` and `settings.gates`. A missing frame or key reads as on (the host enforces). It passes settings frames on to subscribers.
- `room/gif-decode.js` holds the part of `room/gif.js` that does not need three.js, as `decodedSource(url, { maxEdge, maxFps, onFrame })`. `room/gif.js` `animatedSource` wraps it in the CanvasTexture exactly as before.

**Tests**
- `node --test backroom/shared/hypno/tests/loader.test.mjs`: 5 pass. It uses the real bridge on a fake `chrome.webview`, with `tests/probe-station.js`.
- `node backroom/smoke/room-check.mjs`: all room checks pass. The wall screens still play GIFs (10 frames in 1 s, cap 12). Evidence: `C:/wt-br2/_evidence/hypno/k1/room-check-headless/`, log `room-check.log`.

Note: `decodedSource` has one addition to the contract signature: the optional `onFrame` callback that the texture wrapper uses.


CodeBambi:
Fix round on the k1 stack (merged upward through -b..-f, no force push):

- #1164: ctx.onSettings after a station closed subscribes nothing (closed flag on the subs set, also on the superseded catch path). shared/hypno/tests joins ArcademyHarnessExclude (+Abs), checked with `dotnet msbuild -getItem:Content`.
- #1165: webgl is now a real getter. The old `Object.assign` read it once at kit creation, copied a fixed boolean, and made the GL context right then. A draw without `now` reads a per-task clock, so 13 backs in one rAF callback cost 1 render. webglcontextrestored rebuilds the field.
- #1166: the `` still sets `crossOrigin = 'anonymous'` before src. The g0 comment now says the host acks a picture it did not deal as skipped `unknown` (checked against the H1 BackRoomFxPlan).
- #1167: a hold step is idempotent per fx id, so playing roulette.run every frame holds one haze. The table's wash strength wins over the caller's, which removes declared deviation 7. A settings frame below Full, or with reduced motion, releases a haze hold. wheel.turn and roulette.run are documented as played once.
- #1168: sidecar floats are written short (duty 0.3), and `--check-only` still matches a fresh weave.
- #1169: kit-check covers the task clock, reading webgl before any draw, and a lost-then-restored context.

Evidence: C:/wt-br2/_evidence/hypno/k1/fix-round/ (kit-check.log, unit-tests.log 93/93, room-check.log, wheel-check.log, weave-check-only.log, kit-check-headless/).

CodeBambi:

PR #1165, opened 2026-09-14
feat(backroom): hypno kit Loom fields, one shared context (K1 2/6)

CodeBambi:
Lane K1, part 2 of 6. Base: `feat/br2-hypno-k1`.

**`shared/hypno/loom.js`**, built on the real `arcademy/engine/loom/loomField.js` (imported, never forked):
- `LOOM_PRESETS` {backs, hub, whirl, wake, screen}: the v3 mockup presets in schema v2, run through `normalizeParams2` and deep frozen.
- `LOOM_BACKING` {long 512, small 256}; `phaseForAngle(name, rad)` on loomField's span rule; helpers `phaseAt`, `spanRad` and `backingFor`.
- `createLoomKit({still, log})` returns `{webgl, draw, paint, setStill, dispose, debug}`:
  - One GL canvas for the page. It is made on first use and lost (`WEBGL_lose_context`) when the last kit is disposed.
  - The backing store is at most 512 px on the long side, with the aspect rounded to 0.05.
  - The same preset at the same phase and store renders once per frame, so 13 card backs cost one render.
  - With no WebGL, or a lost context: `drawFallbackFrame` into a 2D store of at most 256 px.
- `still` holds phase 0, or the last angle each preset was given.

**Handedness (law 3)**: `tests/handedness.js` measures it on pixels; kit-check runs it in part 6. With WebGL and with the 2D fallback, every preset's layer 1 turns clockwise with a clockwise angle, and its arms lead at the rim (reads inward). One thing for the owner to look at: the gold layer 2 of `screen` turns the other way, as in the approved mockup and the contract preset, so that thin layer on its own reads outward.

**Tests**: `node --test backroom/shared/hypno/tests/loom.test.mjs`: 7 pass.


CodeBambi:

PR #1166, opened 2026-09-14
feat(backroom): hypno kit deck and mock host (K1 3/6)

CodeBambi:
Lane K1, part 3 of 6. Base: `feat/br2-hypno-k1-b`.

**`shared/hypno/media.js`**
- `DECK_VALUES` is A..K with T for ten (the pure module's rank order).
- `createDeck(ctx, {count = 13, maxEdge = 192, still})` asks `ctx.media({count})` and maps value i to `gifs[i % n]`, with no padding. It returns `{seed, keys, size, keyFor, keyAt, pickKey, draw, image, tick, setStill, dispose, debug}`.
- `keyFor('Qh')` and `keyFor('10')` both work. `pickKey(result string)` hashes with FNV-1a, so a result always shows the same picture.
- It draws only URLs on ccp.assets, ccp.game or the page's own origin.
- Decoding goes through `room/gif-decode.js`:
  - At most 13 sources, 192 px, 12 fps, loaded one at a time.
  - At most one decode starts per `tick`, and only sources drawn since the last tick advance.
  - With no decoder, or a file the decoder refuses, it draws a still snapshot of an ``.
- An empty deal (no host, a lost reply) still has one key, `g0`, so `size >= 1`. The host swaps an unknown key for a dealt item.

**`tests/mock-host.js` `createMockHost()`**, for station checks:
- `attach(ctx)` adds `fx`, `fxRelease`, `fxTunnel`, `media` and `onSettings`, plus live `gates`, `intensity`, `reduced` and `motion` getters.
- It records `fx`, `tunnel`, `release` and `media` calls.
- It acks like the host: `toggle` when a gate is off, `calm` for haze below Full, `busy` inside the wash gap.
- It deals by the 10.13.C rule, and `settings(patch)` fires a live settings frame.

**Tests**: `node --test backroom/shared/hypno/tests/media.test.mjs`: 8 pass.

Deviation: `image(key)` returns a canvas in both cases (the decoded frames, or the still snapshot) rather than the `` itself, so a texture never animates on its own.


CodeBambi:

PR #1167, opened 2026-09-14
feat(backroom): hypno kit moments with gates (K1 4/6)

CodeBambi:
Lane K1, part 4 of 6. Base: `feat/br2-hypno-k1-c`.

**`shared/hypno/moments.js`**
- `MOMENTS`: the 15 ids of 10.13.F, frozen, each with its Normal args and page effect names. `velvet_wake` is returned only at Full, and `pulled_pair` only with `{wake: true}`.
- `wheelSize(result)` reads `jackpot`, or `jackpotWon` from the wheel page. Also exported: `strengthK(ctx)` and helpers `wheelTurnLevel(dim)`, `rouletteRunLevel(ballSpeed)`, `pocketColor(pocket, rose)`, `viewportRect(el, x, y, w, h)` and `boxAround(cx, cy, w, h)`.
- `createMoments(ctx, {station})` returns `{play, tunnel, holdScreen, release, cancel, dispose, debug}`.

**`play`**
- Drops every step whose gate is off and fires the rest through `ctx.fx` with Normal args.
- A fixed table colour wins over the caller's `color`. The caller's `strength` replaces the table's.
- `from` is rounded; it is dropped if it is not finite or is under 8 px. `gif` must be `g0` to `g99`.
- Hold tokens (haze, the wake spiral) are released before any `roulette.land.*` fires, and when their gate goes off.
- A `cards.win` right after `cards.bloom` sends the 0.9 wash with no picture.
- `cards.lose` runs the tunnel breath 0.75 x sin(PI p) over 2600 ms.

**Tunnel, hold and cancel**
- `tunnel(level)` posts on a change at most every 100 ms, and re-posts every 1000 ms while the level is above 0.
- `holdScreen(true)` stops host fx and the tunnel.
- `cancel` sends tunnel 0 at once, releases every hold and stops every timer.

**Tests**: `node --test backroom/shared/hypno/tests/moments.test.mjs`: 10 pass, using mock timers for the throttle, keep-alive and breath. All kit node tests: 30 pass.


CodeBambi:

PR #1168, opened 2026-09-14
feat(backroom): woven Back Room spirals (K1 5/6)

CodeBambi:
Lane K1, part 5 of 6. Base: `feat/br2-hypno-k1-d`.

**Woven spirals (10.13.B, spiral source 2)**
- `Scripts/weave-backroom-spirals.mjs` runs the Loom's own encoder, `dtrh/engine/loomWorker.js`, as a module worker in headless Chrome on `LOOM_PRESETS.screen` and `.wake` (format `wide`). It writes `shared/hypno/spirals/{screen,wake}.gif` and their Loom v2 `.json` sidecars.
- `--check-only` weaves again in memory and compares against the files on disk. The weave is deterministic and matched. `--report` prints sizes only. `WEAVE_PORT` defaults to 8894.
- Output:

  | File | Size | Frames | File size |
  |---|---|---|---|
  | `wake.gif` | 640x360 | 63 at 4 cs | 2.74 MB |
  | `screen.gif` | 512x288 | 72 at 5 cs | 5.31 MB |

- kit-check (part 6) decodes both files. From frame 0 to frame 2 each turns clockwise, and its arms lead at the rim.

**Deviations that need a decision**
1. The long side is 640 (wake) and 512 (screen), not 720. The worker's size is fixed at 640, and it drops to 512 once a weave passes its 6 MB soft cap. Getting 720 would mean editing the Loom encoder.
2. `screen.gif` is 5.31 MB, over the 4 MB budget but under the Loom store's own 8 MB cap; the script warns about it. 72 frames of gradient threads plus a second layer do not fit in 4 MB at this size. Options: accept it, lower the preset's frame count (its speed), or let the worker take a size.

Log: `C:/wt-br2/_evidence/hypno/k1/weave.log`.


CodeBambi:

PR #1187, opened 2026-09-14
Back Room: Soft Hand station live in the room [cards 3/5]

CodeBambi:
Soft Hand stack 3 of 5, on `feat/br2-hypno-cards-b`.

## What
- `stations/cards/station.js`: `mount(ctx)` -> `{open, close, suspend, destroy}` through the room loader like the slot and the wheel. A sit-down asks `ctx.media({count: 13})` (through `createDeck`) and plays `cards.sit` (the sit fan). Deal, Hit, Stand, Double, Split from `legal`; bet 1 or 2 SP (1 below 30 SP); basic-strategy hint off by default (`localStorage` `br_cards_hint`); "Stand up, sit back down" re-deals while no hand is open; an open hand resumes after the fan with its decisions live.
- Moments only through `createMoments`, on the frame shown: `cards.bloom` as the blackjack's second card finishes turning (`fx.gif_from` from the ace's viewport rect with the ace's key), `cards.win` / `lose` / `push` on the settle frame (the win wash carries the highest winning card's key). `holdScreen(true)` whenever a reply shows `hand.done === false`, `false` on the settle frame. Deal waits out a bloom (4 s) so it never overlaps the next decision.
- SP through `ctx.spReadout` with Law I; Back live every frame; `close()` synchronous (cancels the moments, disposes the kit and the deck, hands over the plain number). `suspend(true)` lays queued steps down quietly, cancels the moments, frees the Loom context.
- Calm, reduced motion and gates read live every frame (settled steps, still Loom and pictures, page strength x0.5; host args stay Normal).
- Space / Enter deal; moves have no letter keys (the room walks on WASD; a held S or D must never stand or double).
- `stations.json`: the `cards` row goes live (`entry: stations/cards/station.js`); `smoke/room-check.mjs` expects it.

## Tests
`node backroom/smoke/room-check.mjs` passes with the row live (evidence `C:/wt-br2/_evidence/hypno/st-cards/room-check/`, log `C:/wt-br2/_evidence/hypno/st-cards/room-check.log`). The station is driven end to end in 4/5.


CodeBambi:

PR #1202, opened 2026-09-14
Back Room follow-up D-HOST 1/3: room Options, own tunnel vision and melt switches (both on)

CodeBambi:
Lane D-HOST, part 1 of 3 (stacked: this -> feat/br2-hf-host-b -> feat/br2-hf-host-c). Base: the hypno rollup #1201.

## What
Owner decisions after the v3 build (CONTRACT 10.14, written in part 3):

- **Tunnel vision is the room's own switch.** `AppSettings.BackRoomTunnel` (default on) gates `fx-tunnel` instead of Brain Drain. Sent as `gates.tunnel` in `init` and every `settings` frame. Turning it off cancels a running tunnel at once and still pushes the frame (the old Brain Drain branch skipped the frame).
- **Back Room melt on by default.** `AppSettings.BackRoomMelt` (default on) is now the only gate on `fx.melt`. What kept it dark before: the slot's melt needed the app-wide `BrainDrainEnabled` (default off) and lost its drip without `BrainDrainMeltEnabled` (default off). A settings file without the new key reads on (new and existing players); a player's off is kept. `fx.haze` stays on Brain Drain. The app's own Brain Drain feature and defaults are untouched.
- **Room Options card** in the HUD (Options pill beside Room view and Motion): Effects Calm/Normal/Full (the existing setting), Tunnel vision, Melt. Presses post `room-option`; the bridge accepts only `{tunnel|melt: bool}` and `{intensity: calm|normal|full}`; the host writes and saves. `init`/`settings` carry `intensityChoice` so the card shows the player's choice while Calm is forced. Escape/Back closes the card first.
- en.json rows for the ten `br_opt_*` keys.

## Tests
- `dotnet build ConditioningControlPanel.sln` 0 errors; full client tests 6,746 / 6,746 (on the stack tip).
- New `BackRoomRoomOptionsTests` (defaults, old-file read, message shapes, bridge, apply); melt and tunnel gate tests updated.
- `room-check.mjs` 106 ok, 0 fail, new block 2f2 (Options card, room-option posts, host frame wins, Escape). Screenshot `room-options.png`.
- Evidence: `C:/wt-br2/_evidence/hypno-followup/d-host/` (`room-check.log`, `room-check/room-options.png`, `client-build.log`, `client-tests.log`).

## Merge notes
- `room/main.js` `readGates` also changes on the kit lane (fcd322c43, adds `tunnel`). This one is a superset (`tunnel` and `melt`): keep this side.

No desktop app was launched.

[...]

PR #1203, opened 2026-09-14
Back Room follow-up D-HOST 2/3: spiral per screen, flash XP for room pictures, GIF cycling

CodeBambi:
Lane D-HOST, part 2 of 3 (stacked on feat/br2-hf-host).

## What
- **Loom spiral, one field per screen.** `spiral-loom` (and the slot's `spiral-full`, which plays through it) draws a UniformToFill field in a clipped cell per monitor, centred on that monitor, like the tunnel. Cells come from `BackRoomOverlayMath.ScreenCells`: physical monitor bounds through the overlay window's own DPI (the same mixed-DPI rule as the other Back Room overlays). Unchanged cells keep their images; a DPI or display change rebuilds them.
- **XP like the app's own effects.** In normal play only flash images (FlashService, 4 XP per image without its sound, times the lucky flash roll) and subliminals (10 XP) pay; spirals and Brain Drain pay nothing. Back Room `flash-burst` and the sub primitives already go through those services, so they keep paying there and are not paid again. The Hypno pictures that bypass FlashService (`gif-full`, `gif-from`, a `wash` with a picture) now pay 4 x `RollLuckyFlash` through `AddXP(XPSource.Flash)` once each, only when the sink reports the picture showed. Busy-dropped, merged, cancelled or file-less pictures pay nothing. `IBackRoomFxSink.GifFull` and `Wash` return whether a picture showed.
- **GIF shortfall cycles.** `ResolveSymbols` cycles a GIF index past the deal (`gifs[i % n]`); the slot page cycles `gif0..gif3` over its deal. The host already deals fallback art only when the player has no GIFs.

## Tests
- New `BackRoomFxXpTests` (7), `ScreenCells` mixed-DPI tests (2), slot `media.test.mjs` cycling.
- Full client tests 6,746 / 6,746; Back Room node tests 146 / 146; `room-stations-check.mjs` 77 ok, 0 fail.
- Evidence: `C:/wt-br2/_evidence/hypno-followup/d-host/` (`client-tests.log`, `node-tests.log`, `room-stations-check.log`).

## Not verified
The per-screen spiral on real monitors (mixed DPI, dual screen) needs the owner's desk run; no desktop app was launched.


CodeBambi:
Fix round: rebased on the updated #1202, plus 4142c9b0d. The Loom spiral image in each monitor cell is centred with no size of its own (BackRoomSpiralCellTests renders it and fails 3/3 with the old pinning). Picture XP is paid from a shown callback the overlay calls when the picture is actually on. A picture refused during a display change, or one that fails or comes too late, pays nothing. Overlay windows re-read the virtual screen when they wake from idle. Tip c: 6751/6751 tests.

CodeBambi:

PR #1208, opened 2026-09-14
Back Room hypno follow-up: screen spiral layer 2 turns inward, fx-tunnel reads gates.tunnel

CodeBambi:
Hypno follow-up, lane D-KIT, part 1 of 4 (stack: this -> -b -> -c -> -d). 181 changed lines plus one re-woven GIF.

**Screen spiral: layer 2 turns with layer 1 (owner decision 4).**
- `LOOM_PRESETS.screen.layer2.direction` -1 -> 1; CONTRACT 10.13.D line amended.
- `spirals/screen.gif` re-woven with `Scripts/weave-backroom-spirals.mjs` (the Loom encoder): 512x288, 72 frames at 5 cs, 5,693,013 bytes (was 5,567,286), under the Loom store's 8 MB cap (over the old 4 MB budget, which DECISIONS item 3 amends). `wake.gif` re-weaves byte for byte (sha256 unchanged), so it is not in the diff.
- The handedness probe now measures layer 2 alone on its gold thread colour. kit-check asserts it turns clockwise and leads at the rim, on the live kit (+3.5/+3.2/+2.8 deg turn, +5.3/+4.3/+3.8 rim) and on the woven GIF. The same probe on the OLD gif read layer 2 at -4.75/-5/-5.75 deg, so it tells the two apart. Screenshot: `_evidence/hypno-followup/d-kit/smokes/kit-check/loom-handedness.png`.

**Tunnel gate (pairs with the D-HOST lane).**
- Moments gate `fx-tunnel` (wheel last turn, roulette run, cards losing edges) on `gates.tunnel`, no longer `brainDrain`. Haze stays on brainDrain. A host that sends no `tunnel` reads as on, so the current host keeps working.
- The room's readGates, the loader doc, the mock host (sends and applies `tunnel`), the cards/roulette dress, and the dev pages all carry the fifth gate. Gated off, the pages dress plain: no fx-tunnel, and in-station dims stay.
- Tests: moments.test (brainDrain off still tunnels, tunnel off sends no level and no breath, a live settings frame closes it); kit-check room section (an init with no tunnel reads on; settings tunnel:false stops posts).

[...]

PR #1209, opened 2026-09-14
Back Room hypno follow-up: wheel hub turns at half under Calm, reveal count clamped to the pay

CodeBambi:
Hypno follow-up, lane D-KIT, part 2 of 4 (on #1208). 139 changed lines.

**Wheel hub under Calm keeps turning at half strength, as in the mockup.**
- `hypno.stepHub(..., k)` scales the drift and the follow by k (0.5 under Calm or MotionLevel Reduced).
- `hypno.hubStill`: only OS `prefers-reduced-motion` holds the hub still, plus the app's Motion Off (see the note below).
- wheel-check measures the drift: 0.350 rad/s at Full, about 0.175 under Calm and under reduced. With OS reduced motion emulated through CDP, the hub shows 0 deg on screen. A live settings frame is covered for reduced (still turning) and off (held). New shots: st-wheel-19 (Calm hub turning), st-wheel-20 (OS reduced hub still).
- Note: MotionLevel Off also holds the hub still. It is the player's explicit no-motion choice, and dropping it is a one-line change in `hubStill`.

**The count-up never shows more than the pay.**
- `feel.revealCount(pay, q)` clamps THE REVEAL count to 0..pay (the motion ease still overshoots). This fixes "JACKPOT +551" for +550.
- Node test sweeps 201 steps for five pays. wheel-check samples the status screen every frame of the reveal: 13 values, max 550.

**wheel-check at-speed hub flake.** Two captures could straddle 200+ deg of hub turn and alias past the probe's +-60 deg window. The integration run hit this as a flake. The probe now reads a pair only when the hub's own angle turned 1..55 deg between the captures, and tries up to five pairs. A backward hub still fails every readable pair.

Verification is on the stack tip; see #1208.


CodeBambi:

PR #1211, opened 2026-09-14
Back Room hypno follow-up: en.json rows for the station br_* keys

CodeBambi:
Hypno follow-up, lane D-KIT, part 4 of 4 (on part 3). 234 changed lines.

**Lexicon: en.json rows for every br_* key the stations use.**
- 141 rows after `br_wheel_slice_dazed`, each one the page's inline English:
  - `br_wheel_slowly`
  - 39 `br_slot_*`, including the odds line labels and the gain readout
  - 49 `br_cards_*`, including the four moves, the net lines and the new `br_cards_moment`
  - 52 `br_roulette_*`, including the rows, spots and mat labels
- Cards and roulette follow their station.md order; the slot follows code order. The integration's 46/46/31 counted static calls only; keys built at runtime are included here too.
- Other languages fall back to English through LocalizationManager (`KeysWithPrefix` reads the English file). Nothing is machine translated.
- `smoke/lexicon.test.mjs` scans the page sources. It fails on a key with no row, or on a row that reads differently from the page.

Verification (stack tip): node 152/152 including the lexicon test; dotnet build 0 errors; client tests 6,720/6,720 (BackRoomHypnoBridgeTests read en.json br_ keys).

## Deal lock (2026-09-14, playbook-join, commit 8566eac12)

Owner: "they shouldn't be able to click deal when waiting, nip it in the bud". While a fullscreen moment runs (bloom, win flash, losing edges) Deal cannot be pressed by any path; presses are dropped, never queued.
- `deal()` refuses first (no ring, no note, a `dropped` counter in debug), so Space, Enter, repeated or scripted clicks and a direct call drop inside the deal action itself.
- `hand.controls` reads `screen` before `busy`, so the button is disabled, marked `data-held` and reads "One moment" from the frame the bloom fires, even while the hand is still being laid down.
- `moments.breathing()`: a tunnel breath that outlives its timed hold keeps the hold.
- cards-check 1b, during a bloom: four real CDP mouse clicks, Space and Enter on the table and on the focused button, six scripted clicks, a direct `onclick()` and a click on a re-enabled button send no deal (7 presses dropped); after the moment no deal follows on its own and one click deals once; `deal()` has exactly two callers and no gamepad, room or host message reaches it. cards-check 84 ok, 0 fail; cards node tests pass.


CodeBambi:

PR #1231, opened 2026-09-14
feat(prizes): profile sync and validate feed App.Ownership (H-own)

CodeBambi:
## What

Back Room CONTRACT 10.17.E, feed (1): profile sync and the provider validate responses feed `App.Ownership`.

- `/v2/user/sync`: the `prizes` block `{ revision, grants }` goes to `OwnershipService.ApplySnapshot` for the `unified_id` the sync was sent for. Applied right after the response is parsed, before the other flag handling, so a later throw cannot skip it.
- `/patreon/validate`, `/discord/validate`, `/substar/validate`: the same block (CCP-Server S-reach) is applied only when the response's `unified_id` equals the session's unified id captured before the request. A duplicate-identity pair (validate resolving to another record) is ignored, same rule as the auth-token heal.
- Missing block (older server, or `grants` absent): held state is left alone. An empty `grants` list is a real snapshot.
- Logout: `ResetLoadedProfileState` (every logout path funnels through it) clears ownership.
- Account switch: the V2 sync notes the account before the request; a different account than last time clears first, so listeners hear the removal before the next account's snapshot lands. `OwnershipService` already hides a previous account's grants the moment `App.UnifiedUserId` changes.
- Grants are never persisted (not on `PatreonCachedState` / `DiscordCachedState`, not in settings) and never uploaded.

The logic sits in `ProfileSyncService.PrizesFeed` (injected ownership source) so it tests without App or a dispatcher. DTO additions: `PrizesBlock` (both serializers' attributes) plus a `Prizes` property on the two validate responses and the private `V2SyncResponse`.

## Tests

- `dotnet test --filter ProfileSyncPrizesFeedTests|PrizeOwnershipServiceTests`: 37 passed (9 new)
- Full suite `dotnet test`: 6282 passed, 0 failed
- New cases: snapshot applied, missing block leaves state alone, empty grants applies, other-account response ignored, response for an account no longer signed in ignored, logout clears, account switch clears, block parses with System.Text.Json and Newtonsoft, grants never on cached provider state.

Not run in the app (headless only). Base: `feat/br2-ownership-api` (H0).

https://claude.ai/code/session_01GkR5D99oah2WMcLV991yAd


CodeBambi:
Landed on main via #1234.

PR #1251, opened 2026-09-14
fix(backroom): casino stack review fixes (parts 1-6)

CodeBambi:
Five opus reviews of the six merged casino parts (#1245-#1250), fixed on the integrated code. Web only, except one csproj exclude.

### From #1246 (EMI)
1. **`room/emi-idle.js` `update()`** - the still branch now clears the action (`action = null; age = 0;`) before `rest()`, so a gesture caught by Motion off / Calm / reduced is dropped instead of parked to replay whole when motion returns.
2. **`room/emi-interaction.js` + `room/scene.js`** - a new `canGesture()` option, passed as `() => !still`, so a click on a still room shows the bark bubble and skips the gesture.
3. **`room/tests/emi-gestures.test.mjs`** - a bound test for the IDLE path is back. The old ceiling (yaw .085, pitch .05, roll .016) no longer describes it: the idle rotation plays the authored reactions, so `look` sweeps the head .34 rad and `wave` lifts a shoulder. The bounds are set to what idle really needs (yaw .34, pitch .16, roll .045, `lift === 0`, face null or the rest face 3) over 3 periods on every station, with the reason written into the test. `present` stays in the rotation. Click-driven reactions (`sampleEmiReaction`) stay free.
4. **`room/scene.js` `halt()`** - `emi.dispose()` for every `room.emis` entry, next to the decor and interaction disposal. The "Review copy only" header in `emi-interaction.js` is gone (the sentence about barks implying no outcome is kept).
5. **`room/tests/emi-idle.test.mjs`** - a new test that a gesture caught by Motion off is dropped, never replayed, and that a fresh click still lands.

### From #1247 (celebrations, props)
6. **`stations/slot/scene.js` coin shower** - takes a `stillFx()` flag from `station.js` that is true for reduced OR `ctx.intensity === 'calm'` OR `ctx.motion === 'off'`, read live so a settings frame applies mid sit-down. It was passing `reduced` alone.
7. **`room/fixtures.js` shell deform** - the replaced geometry is disposed after the swap. The `max.x <= 6.5` skip and the kept exported normals for skipped meshes were already in the merged code.

### From #1248 (handles)
8. **`room/customization.js`** - `if (!holder) return;` before `getWorldPosition` in the handle preview path.
9. **`stations/slot/nodes.js` + CONTRACT.md section 6** - `wand_head` and `wand_switch` added to `OPTIONAL`, and to the section 6 node list with a line of purpose.
[...]

PR #1250, opened 2026-09-14
Back Room 6/6: Ceiling projector, final layout and performance budgets

CodeBambi:
Projects rotating media across the whole ceiling and adds a miniature projector to the vending machine. Enlarges the vending machine, spreads statue pedestals, adjusts roulette spacing, removes the selection outline and tunes walking feel. Makes room slots 10% larger with matching bounds.

Bounds rendering resolution/frame rate, stops hidden-room rendering, shares sculpture resources and limits media decoding. At the tested entrance view: 492 to 249 draw calls, 1,132,268 to 566,728 submitted triangles, and cold response bodies 12.10 MB to 9.94 MB. Customization GLBs shrink from 3.12 MB to 0.96 MB. PERFORMANCE.md records methodology and website caching requirements. Physical low-end phone and WPF host validation remain outstanding.

Part 6/6 of the casino review stack. Review and merge bottom-up; retarget dependents after merging their base. Original approved prize models are already merged in #1216.

Validation: 120 room/slot Node tests passed on the final integrated stack. Full room browser smoke passed on the final integrated stack with no page errors, including live-counter registry integration, movement/collisions, EMI barks, station hold/resume, GIF caps and bell request checks. Earlier local checks covered projector timing, desktop/mobile pixel budgets, zero-download lever swaps and all six lever/sculpture pose combinations.

Draft review only. No deployment.

## Review order

1. [Room layout, lighting and casino decoration](https://github.com/CodeBambi/Conditioning-Control-Panel---CSharp-WPF/pull/1245)
2. [Articulated EMI animations and clickable barks](https://github.com/CodeBambi/Conditioning-Control-Panel---CSharp-WPF/pull/1246)
3. [Payout celebrations and Room Service models](https://github.com/CodeBambi/Conditioning-Control-Panel---CSharp-WPF/pull/1247)
4. [Chess sculpture levers and room customization](https://github.com/CodeBambi/Conditioning-Control-Panel---CSharp-WPF/pull/1248)
5. [Physical vending showcase and media upgrades](https://github.com/CodeBambi/Conditioning-Control-Panel---CSharp-WPF/pull/1249)
6. [Ceiling projector, final layout and performance budgets](https://github.com/CodeBambi/Conditioning-Control-Panel---CSharp-WPF/pull/1250)

PR #1249, opened 2026-09-14
Back Room 5/6: Physical vending showcase and media upgrades

CodeBambi:
Uses the actual vending machine for selection in the right third of the screen, with the room preview in the left two thirds. Adds clickable 3D spiral samples, independent four/six extra-screen packs and full-sculpture levers. Retires non-palm plants from the active catalogue.

Part 5/6 of the casino review stack. Review and merge bottom-up; retarget dependents after merging their base. Original approved prize models are already merged in #1216.

Validation: 120 room/slot Node tests passed on the final integrated stack. Full room browser smoke passed on the final integrated stack with no page errors, including live-counter registry integration, movement/collisions, EMI barks, station hold/resume, GIF caps and bell request checks. Earlier local checks covered projector timing, desktop/mobile pixel budgets, zero-download lever swaps and all six lever/sculpture pose combinations.

Draft review only. No deployment.

## Review order

1. [Room layout, lighting and casino decoration](https://github.com/CodeBambi/Conditioning-Control-Panel---CSharp-WPF/pull/1245)
2. [Articulated EMI animations and clickable barks](https://github.com/CodeBambi/Conditioning-Control-Panel---CSharp-WPF/pull/1246)
3. [Payout celebrations and Room Service models](https://github.com/CodeBambi/Conditioning-Control-Panel---CSharp-WPF/pull/1247)
4. [Chess sculpture levers and room customization](https://github.com/CodeBambi/Conditioning-Control-Panel---CSharp-WPF/pull/1248)
5. [Physical vending showcase and media upgrades](https://github.com/CodeBambi/Conditioning-Control-Panel---CSharp-WPF/pull/1249)
6. [Ceiling projector, final layout and performance budgets](https://github.com/CodeBambi/Conditioning-Control-Panel---CSharp-WPF/pull/1250)

PR #1245, opened 2026-09-14
Back Room 1/6: Room layout, lighting and casino decoration

CodeBambi:
Adds the decorated Prize Parlour sign, animated booth lights, shared EMI idles, chandeliers and palms. Preserves the approved left-wall game placement, restores strafing and fits the east screen. Current main's live prize counter is retained.

Part 1/6 of the casino review stack. Review and merge bottom-up; retarget dependents after merging their base. Original approved prize models are already merged in #1216.

Validation: 120 room/slot Node tests passed on the final integrated stack. Full room browser smoke passed on the final integrated stack with no page errors, including live-counter registry integration, movement/collisions, EMI barks, station hold/resume, GIF caps and bell request checks. Earlier local checks covered projector timing, desktop/mobile pixel budgets, zero-download lever swaps and all six lever/sculpture pose combinations.

Draft review only. No deployment.

## Review order

1. [Room layout, lighting and casino decoration](https://github.com/CodeBambi/Conditioning-Control-Panel---CSharp-WPF/pull/1245)
2. [Articulated EMI animations and clickable barks](https://github.com/CodeBambi/Conditioning-Control-Panel---CSharp-WPF/pull/1246)
3. [Payout celebrations and Room Service models](https://github.com/CodeBambi/Conditioning-Control-Panel---CSharp-WPF/pull/1247)
4. [Chess sculpture levers and room customization](https://github.com/CodeBambi/Conditioning-Control-Panel---CSharp-WPF/pull/1248)
5. [Physical vending showcase and media upgrades](https://github.com/CodeBambi/Conditioning-Control-Panel---CSharp-WPF/pull/1249)
6. [Ceiling projector, final layout and performance budgets](https://github.com/CodeBambi/Conditioning-Control-Panel---CSharp-WPF/pull/1250)

PR #1252, opened 2026-09-15
fix(backroom): casino follow-ups: phone panel layout, orphaned br_custom keys, CONTRACT drift

CodeBambi:
Three follow-ups the opus reviewers raised on the casino stack (#1245-#1250) that the #1251 fix-up deliberately left undone. Web only, no C# change.

### 1. Room Service on a phone

`room/customization-panel-style.js` sized the panel at a fixed `33.333333vw`. On the 400 px viewport the room is played at through cclabs-web that is about 133 px, which is unusable.

- **Breakpoint: 700 px**, the panel's own existing one (it already shrank its header, item grid, HUD and footer there). The room's stations use 800 px; the panel had 700 and reusing it keeps one breakpoint rather than two.
- Under it the panel is a **full-width sheet on the bottom of the screen**: `left:0; width:auto; top:auto; height:76vh`, border on the top instead of the left, and `.br-custom-hud` capped at `30vh` instead of `40vh` so the stage keeps a rectangle worth drawing into (measured 400 x 379).
- The **#1251 item 10 offsets are respected by being dropped where they no longer apply**: with no panel edge to sit left of, `.br-nav` and `.br-bell` go back to `right:12px` and the bell to `calc(100vw - 24px)`, in their walking band at the top. They are still never hidden.
- The vending close-up (#1251 item 11) needed nothing: it reads the stage rect from the DOM each draw, so it follows the new shape by itself.
- **The ROOM's own pass did need a fix**, and this was a real bug the phone layout exposed: `room/scene.js` hardcoded `previewWidth = fullWidth * 2/3`, so the walking room drew into a 266 px strip of a 400 px phone with the rest of the canvas stale. It now asks `customization.previewBox(w, h)` for what the panel leaves it, the strip beside the panel or the band above the sheet, and sets its viewport, scissor and camera aspect from that. At 1280x720 the box is the same 853 px strip as before, so nothing moves on the desktop.

`smoke/room-check.mjs` section 2k still passes unchanged, and a new phone pass (2k2) boots at 400x800, opens the panel and asserts the panel width fits the viewport (400 of 400), the sheet starts below the chrome, the Options pill row and the floor bell are wholly on screen, the pill is still the element under its own pixels, the stage keeps a usable rect, the close-up follows it, the room keeps drawing, no page errors, and `document.querySelectorAll('canvas').length` is 1 throughout. Screenshot `room-service-panel-400x800.png`; it reboots to 1280x720 afterwards.

### 2. Orphaned `br_custom_*` rows

Sixteen `br_custom_` rows nothing reads, including the six GLBs #1250 deleted, removed from **all nine** `Localization/Languages/*.json` (key order unchanged, every file still parses):

`br_custom_preview`, `_categories`, `_unavailable`, `_displays`, `_plants`, `_gallery`, `_portraits`, `_billboard`, `_monstera`, `_ivy`, `_terrarium`, `_live`, `_none`, `_decor`, `_handles`, `_floor`

Four more the new reverse check turned up alongside them, dead since the commit that added them and outside every family: `br_soon_title`, `br_station_scratcher`, `br_model_missing`, `br_suspended`. Twenty rows in total, 180 lines across the nine files. Nothing that is still read was touched.

**How the lexicon reaches the page:** `BackRoomHostService` sends every `br_` row from the current locale in `init.lex` (`LexPrefix = "br_"`), so there is no web-side lexicon file; a page either spells a key out, or builds one. `smoke/lexicon.test.mjs` gained a fourth test that mirrors that: every `br_` row in `en.json` must be named outright by a page source, by `stations.json` or by a `Services/BackRoom/*.cs` file, or be built by one of four declared families, each of which says where its suffixes come from:

| family | suffixes read from |
| --- | --- |
| `br_custom_` | the panel's own item and control tables in `room/customization-panel.js` |
| `br_emi_` | `stations.json` ids x the three line numbers |
| `br_wheel_slice_` | the wheel's slice ids, without their `_a` / `_b` half |
| `br_station_` | `stations.json` ids |

It reported exactly these twenty rows before the deletion and reports none after. No allowlist.

### 3. CONTRACT.md drift

- **Section 7, wall screens.** "go on the four screens" became the room's own four (`media_screen_0..3`) plus whatever Room Service has switched on, which is two further packs of wall screens, four and six, and the ceiling projection (#1249/#1250). The `room/screens.js` header carried the same stale line and was corrected with it.
- **Section 10.16.A, the spiral jar.** "The room would need ... a new `ctx` channel; neither is cheap enough to be worth it" is no longer true: `ctx.revealedWin(amount, tier, text)` exists and the coin shower spends it. The passage now says the channel is there, what it carries (the pay, its tier, the win text, never the jar's fill) and why the jar is still not drawn from the room.
- Grepped for other counts the stack touched: `decor`, statues, pedestals and the deleted GLB names appear nowhere in the CONTRACT (Room Service has no section of its own), and the `media_screen_0..3` node list is still correct. Nothing else changed.

### Verification

From `Resources/web/backroom`:

- `node --test room/tests/*.test.mjs stations/{slot,wheel,cards,roulette,counter}/tests/*.test.mjs shared/hypno/tests/*.test.mjs` - **280 tests, 0 fail**
- `node --test smoke/lexicon.test.mjs` - **4 tests, 0 fail** (1 new)
- `node smoke/room-check.mjs` - **all room checks passed**, 14 new assertions in the phone pass, screenshot eyeballed: the room band draws full width at the top, the sheet fills the rest and the vending close-up fills the stage
- `node smoke/room-stations-check.mjs` - **all room station checks passed**

From the worktree root:

- all 9 `Localization/Languages/*.json` parse
- `dotnet build ConditioningControlPanel.sln -c Debug` - **0 errors** (463 pre-existing warnings)
- no em-dashes


CodeBambi:
Opus review: MERGE. Desktop 1280x720 previewBox byte-identical to the old 2/3 strip (probe: 853 px), the 4 extra removed keys have zero references in client, xaml, web bundle and cclabs-web, nav/bell never hidden, close-up reads its rect from the DOM. Lexicon 4/4, room-check and room-stations-check green, room tests 31/31. Nit fixed on the branch: ceiling projection turns every 4.5 s.

PR #1253, opened 2026-09-15
feat(flashes): Shatter, a v2 dismiss that breaks the gif into falling shards

CodeBambi:
Flashes v2 wave 2, third effect, after Rounded corners and Draggable gifs.

## Design

Dismissing a flash used to be a plain cut: the picture was simply gone the instant the click landed. Now it breaks.

- **The cut.** `FlashShatter.Create` splits the picture along a 3x2 or 3x3 grid, rolled per dismiss, so a break is **6 or 9 shards**. Every interior cut line is jittered up to 12% of a cell off the even split, so no two breaks of the same gif produce the same pieces, and the outer edges stay put so the shards still tile the picture exactly once. The cut is normalised (0..1 across the media box), so the maths never has to know the picture's pixel size or the monitor's dpi.
- **The fall.** Each shard takes an outward kick away from the middle of the picture (90..220 px/s, so the break opens rather than sliding) plus a downward shove (30..140 px/s), then falls under **1400 px/s^2**, turning up to 2.4 rad/s about its own centre. A shard tumbles a little; it does not spin like a coin.
- **The fade.** Full opacity for the first 30% of the break so the eye registers the pieces, then straight down to nothing at **0.70 s**.
- **Done** when the fade runs out or every shard has left the monitor, whichever comes first, so a flash broken on the bottom edge ends early instead of animating pieces nobody can see.
- **Motion Reduced** breaks into a flat **4** over **half the duration**. **Motion Off** returns a done, shardless state, which is the signal to keep the plain cut.

**Compositor path only**, exactly like dragging and for the same reason: a classic per-window layered flash deadlocks on `CompleteRender`, so on that path the dismiss stays a cut.

**Which dismiss counts:** `OnFlashClicked` and nothing else. That is the tap `FlashDrag` already routes there, the right-click escape hatch, and the gaze dwell that is documented as the same pop. A fling release never reaches it, and neither does a timer expiry, a retired one-shot or the run stopping, so all of those still fade or cut exactly as before. **XP, hydra and the active list are untouched** - the window leaves the active list, the XP event fires and the hydra roll runs at the dismiss as they always did, and only the picture's exit changed.

**Rounded corners** are deliberately not carried onto the pieces: a break exposes hard edges, and rounding every shard would read as a bag of lozenges.

**Ownership:** `FlashShatterEnabled` defaults **ON** and rides the same Flashes v2 motion grants as rounded corners and draggable gifs (`PrizeGrants.FlashDriftBounce` / `FlashPendulum`, never a setting). It gets a **Shatter** row in the existing v2 box next to Motion, Rounded, Draggable and Remix, so an account that bought none of it sees nothing new. On by default because, unlike the other two dials, it changes nothing until a hand actually dismisses a flash.

## Files

| File | What |
| --- | --- |
| `ConditioningControlPanel/Services/Flash/FlashShatter.cs` | **new** - the pure maths: `FlashShard`, `FlashShatterState`, `Create` / `Step` / `AlphaAt` / `OffScreen` / `ShardCount`. No WPF, no Skia, no App. |
| `ConditioningControlPanel/Services/Compositor/FlashLayer.cs` | `FlashItem.Shatter`, `BeginShatter` (the layer takes the item and its frames over at the dismiss), the break stepped and the item dropped on the **existing** flash tick, and `DrawShards` drawing each piece as a sub-rect of the last frame. No second timer, no new surface. |
| `ConditioningControlPanel/Services/Flash/FlashService.cs` | `FlashWindow.ShatterOnDismiss` set by `OnFlashClicked`, and `BuildShatter` in the compositor teardown branch deciding between `BeginShatter` and the plain `Remove`. |
| `ConditioningControlPanel/Models/AppSettings.cs` | `FlashShatterEnabled`, default true. |
| `ConditioningControlPanel/Features/FlashFeatureControl.xaml(.cs)` | the Shatter toggle row in the v2 box, hidden on the same grants as Rounded and Draggable. |
| `Localization/Languages/*.json` (9) | `setting_flash_shatter`, `tooltip_flash_shatter` - all nine, matching how the Rounded and Draggable keys landed. |
| `Tests/ConditioningControlPanel.Tests/FlashShatterTests.cs` | **new** - 23 tests. |

## Tests

`dotnet build ConditioningControlPanel.sln -c Debug`: **0 errors**.
`dotnet test ConditioningControlPanel.sln -c Debug`: **7137 passed, 0 failed, 1 skipped (7138 total)**, of which **23 new** in `FlashShatterTests` (baseline on main was 7115).

Covered: shard count at each motion level (6 or 9 / 4 / 0), that both Full grids actually get rolled, that Off arrives already done and never moves, that the shards tile the whole picture with no gap and no overlap and the outer edges stay put, that the jitter makes two breaks differ, gravity accelerating every piece downward, the outward opening on the first tick, the tumble going both ways, a zero delta moving nothing, the fade curve and its monotonicity, the ~0.7 s duration, Reduced halving it, and both done conditions (fade out, and every piece off the monitor) plus the no-bounds case.

## Desk test

```
1. DEBUG build with CCP_PRIZE_GRANTS=fx.* so the Flashes v2 box shows with no Back Room purchase.
2. Studio -> Flashes: v2 box has a "Shatter" row, on by default; Solid mode OFF (compositor on), Clickable flashes ON.
3. Start the run and click a flash: 6 or 9 shards fall, tumble and fade in about 0.7s; the XP popup is unchanged.
4. Settings -> Motion = Reduced, click a flash: 4 shards, about half as long. Motion = Off: plain cut, no shards.
5. Let a flash time out untouched (still a quiet fade), then turn the Shatter row off and click one (plain cut).
```


CodeBambi:
Opus review: MERGE at 9df13b624. Blockers fixed: pendulum flashes now break over the pivot-space media rect with a frozen tilt (TakeOverHangingRect), BOM stripped. Notes done: dismiss flag reset above the layer branch, empty-frames guard, ShatterBounds from the monitor under the picture, DwellScale applied, per-shard black backing for non-glow, tooltip copy. 0 build errors, FlashShatterTests 27/27, solution 7141 green. Desk run owed (Pendulum picker in particular).

PR #1254, opened 2026-09-15
feat(bubbles): Magnet bubble, a v2 kind that homes on the cursor and pays double when popped early

CodeBambi:
Bubbles v2, wave 2, second kind. A steel-blue bubble that **steers toward the mouse pointer** as it rises, and pays **double XP** if you take it inside the first 35% of its life. The bargain: it is coming to you anyway, so the points are for taking it now.

## Design

**Where it sits.** Built exactly like the Brain Drain bubble beside it. A pure decision/physics class in `Services/Chaos/` with no WPF, a spec built in `BuildTriggerSpec` rather than a row in `ChaosBubbleVariants.All` (so it never leaks into the Rabbit Hole pool or the chaos setup window), and an id appended to the trigger roll **live** rather than stored in `BubbleTriggerVariants`. Three gates, all of which must pass:

1. Bubbles v2 owned - Rain or Spiral In, read from `PrizeGrants` through `V2Badges.BubbleOwned()`, never from settings, so a synced profile cannot grant itself the bubble
2. the new `BubbleMagnetEnabled` row, default **on** so a bought prize shows up without a second opt-in
3. the user's existing `BubbleTriggersEnabled` switch and their `BubbleTriggerChance` dial, unchanged

It chains after `BrainDrainBubble.RollPool`, so a user who owns v2 and leaves both rows on rolls from their picked ids plus exactly two extras, each equally weighted.

**Attraction.** A capped turn plus a capped speed, never a lerp toward the cursor point: a lerp teleports when the cursor jumps and ends up welding the bubble under the pointer where it cannot be missed. A bubble that can only bend its heading a little each frame visibly hunts, overshoots and comes round again.

| tunable | value | why |
| --- | --- | --- |
| `MaxTurnRadPerFrame` | 0.06 rad | ~1.8 rad/s at 30 fps, a bit under a third of a turn a second |
| `AccelDipPerFrame` | 0.12 | a plain bubble drifts 1..2 DIP/frame, so the ramp takes about a second |
| `MaxSpeedDipPerFrame` | 3.2 | ~96 DIP/s. A safety rail, not an attraction knob, so it does not scale with motion level |
| `MinSpeedDipPerFrame` | 0.35 | a bubble turned to a standstill still has a heading to bend |
| `DeadZoneDip` | 28 | it arrives, it does not sit under the pointer jittering |
| `CursorStaleSeconds` | 2.0 | a pointer parked longer than this is treated as absent |
| `EarlyWindowFrac` | 0.35 | of a 7000 ms trigger life, the opening ~2.45 s |

**Motion level.** Full pulls at 1.0, Reduced at 0.5 (half the turn, half the ramp), Off at 0.0. At Off the bubble is a plain float **and stops paying double**, so switching motion off can never be the cheap way to farm the prize.

**Cursor.** One `GetCursorPos` a tick, taken on the ambient path (the existing sample is chaos-only) and **only while a magnet is actually on screen**, so a user who owns nothing never pays for a P/Invoke. Deliberately a poll and not the `GlobalMouseHook`: a cosmetic bubble drift is nowhere near worth arming a global app-wide input hook, and the poll answers the question the physics actually asks. "Off all monitors" is read from the `Screen` list rather than trusted from the call, because `GetCursorPos` still returns the last good point across a disconnected or mid-transition display.

**Multi-monitor.** A bubble lives on the screen it spawned on and is destroyed the moment it crosses that screen's bounds, so a magnet may only steer at a cursor inside its **own** box - `[_screenLeft, _screenRight + _size] x [_screenTop, _screenBottom]` in DIPs (`_screenRight` is the largest legal top-left X, so the sprite size goes back on to reach the real right edge). Without that bound, a pointer parked on monitor 2 sent every magnet on monitor 1 beelining at the shared edge, across it, and to a silent unpopped death. `CursorUsable` answers "is this sample live at all" desktop-wide, `CursorInBounds` answers "can this bubble reach it", and `CanSteer` is the AND the per-frame branch actually asks.

**Fallback.** No attraction (Off / cursor gone / cursor parked past 2 s / cursor on another monitor) and it falls straight through to the ordinary motion switch, flying whatever row the dashboard's Motion setting gave it. The velocity is re-seeded from that row each frame so the first steered frame after the cursor comes back has a real heading to bend.

**XP.** One path, not two: `AmbientXpMultiplier` is a multiplier applied inside `AwardAmbientPop`, so an early magnet pop still takes the same lucky roll and still draws from the same 300/day ambient bucket. It doubles the rate, never the ceiling.

**Visual.** Magnetic steel blue `#2F6FBF` body (explicitly not the Brain Drain violet `#6A22A8`) with a lighter `#6FA8FF` halo, pulsing at depth 0.12 / 4.4 rad/s - quicker and shallower than the drain's 0.18 / 1.9 breathe, which is how the two tell themselves apart at a glance. The halo is a drop shadow on the bubble element, so riding the opacity the frame already computes pulses the ring with it: **no new element in either render path, no new window types**, and the compositor takes the same colour through `BuildLayerItem`. No sprite ships for it, so it wears the tinted `bubble.png` with a `⊕` glyph (U+2295), kept as a constant on the pure class and guarded by a test against every label in the chaos variant table - the first revision used `◎`, which is the chaos `spiral` variant's.

**No payload at all.** Its whole content is how it moves and what the pop is worth. That also keeps the companion easter egg off it (`IsAmbientEffectBubble` wants a payload) - an avatar pop would collect the early-window bonus the user was supposed to earn.

As a consequence the concurrent-trigger ceiling counts `IsAmbientTriggerBubble || IsAmbientEffectBubble` - the new ambient-trigger flag **OR** the old "carries a payload" predicate, a strict superset and never a swap. The first revision of this PR swapped them, which was a regression the review caught: `_bubbles` also holds chaos bubbles, and a chaos benign treat satisfies the old predicate but is built with `ambientTrigger:false`, so during a chaos run the cap would have stopped counting the chaos field and allowed up to four extra layered windows at the heaviest moment - the exact desktop-heap starvation the cap exists for (#448/#431).

## Files

| file | change |
| --- | --- |
| `ConditioningControlPanel/Services/Chaos/MagnetBubble.cs` | **new.** The whole decision surface and all the physics. Pure, no WPF |
| `ConditioningControlPanel/Services/BubbleService.cs` | roll pool, `BuildTriggerSpec` arm, the steering branch in `AnimateFrame`, `SampleMagnetCursor`, the XP multiplier, the glow in both render paths, trigger-ceiling count |
| `ConditioningControlPanel/Services/Chaos/ChaosBubbleVariants.cs` | a `"magnet_pull"` case in `PopWordFor` -> `PULL` |
| `ConditioningControlPanel/Models/AppSettings.cs` | `BubbleMagnetEnabled`, default true |
| `ConditioningControlPanel/Features/BubblePopFeatureControl.xaml{,.cs}` | the "Magnet bubble" row in the existing v2 box, same styling as the Brain Drain row. The box's own visibility and its champagne pill are untouched - `V2Badges` still owns that rule |
| `ConditioningControlPanel/Localization/Languages/*.json` (9) | `setting_bubble_magnet`, `tooltip_bubble_magnet`. English into all nine, matching exactly what the Brain Drain rows did |
| `Tests/ConditioningControlPanel.Tests/MagnetBubbleTests.cs` | **new.** 66 tests |

Nothing under `Services/Flash`, `FlashFeatureControl` or `Resources/web/backroom` is touched.

## Tests

**66 new**, covering every item asked for:

- attraction turns toward the cursor (heading error closes, and it moves toward rather than away)
- it homes but never teleports: 60 frames from across the screen, no frame past the cap, aimed and closing by the end
- turn rate capped per frame (cursor directly behind, a half turn away, turns exactly the cap)
- speed cap holds over 400 frames, and actually binds
- Reduced halves the turn the same frame would have made; the time scale halves it too
- Off = no attraction at all (byte-for-byte the same velocity) and no double XP
- early-window boundary at 35%: 2449 ms early, 2450 ms not
- cursor-stale fallback: 2.0 s usable, 2.01 s not, off-screen never
- the multi-monitor bound: a 9-case corner table, a live freshly-moved cursor on monitor 2 refusing to steer a monitor-1 bubble, and `CanSteer` as the AND of both rules
- the body glyph collides with nothing in the variant table, and the pop word is wired
- dead zone, stalled-bubble seeding, ring-pulse band, tint distinct from Brain Drain
- pool contains Magnet only when owned **and** enabled, never double-adds, chains cleanly with the Brain Drain pool
- the persisted default is on, an older settings file without the key reads as on, off round-trips

```
dotnet build ConditioningControlPanel.sln -c Debug   ->  0 errors
dotnet test  --filter MagnetBubbleTests              ->  66 passed
dotnet test  Tests/ConditioningControlPanel.Tests    ->  7180 passed, 1 skipped, 0 failed  (7181 total)
```

(`UiThreadStackCaptureTests.AnUnknownThreadIdDegradesInsteadOfThrowing` flaked once on an intermediate run and passed on re-run; it is unrelated to this lane.)

## Desk test

1. Launch with `set CCP_PRIZE_GRANTS=fx.bubble.rain` (any Bubbles v2 grant will do) so the v2 box appears.
2. Settings -> Bubble Pop -> the champagne **Bubbles v2** box: confirm the new **Magnet bubble** row is there under Brain Drain bubble, on by default.
3. Same page, turn **Trigger Bubbles** on and push **Trigger chance** to 100%, then start the engine so ambient bubbles spawn.
4. Watch for a steel-blue bubble with a pulsing halo and a `◎` label, and park the mouse somewhere across the screen: it should visibly bend toward the pointer, overshoot, come round again, and never snap onto it or sit welded under it.
5. Pop one in its first ~2.5 seconds (35% of its 7 s life) and check the XP flourish pays 10 instead of 5; pop a later one and confirm it pays 5. Then set Motion to Reduced (half the pull) and Off (no pull, normal XP), leave the mouse still for 3 seconds to see one give up and go back to a plain float, and **if you have a second monitor**, park the pointer there and confirm the magnets on the other screen float straight up instead of running at the shared edge.

## Known, deliberately left alone

- **The chaos auto-pop surface.** A magnet caught by one of the chaos auto-pop boons would take the early-window bonus without the user aiming at it. Pre-existing for every scoring bubble, and bounded by the same 300/day ambient cap, so it is not worth a special case in this lane.
- **`Preset.cs`.** `BubbleMagnetEnabled` is not carried in presets, exactly as `BubbleBrainDrainEnabled` is not. Consistent with its neighbour; changing it is a separate decision about both rows.


CodeBambi:
Opus review: MERGE at f65783d80 (+ BOM strip). Blockers fixed: magnets steer only at a cursor inside their own screen box (CanSteer = CursorUsable AND CursorInBounds), and the MAX_TRIGGER_WINDOWS ceiling counts IsAmbientTriggerBubble OR IsAmbientEffectBubble so chaos bubbles still count. Notes done: glyph U+2295 with a collision test, PULL pop word, allocation-free cursor sample, MotionFx read once per tick, no free 2 s on a parked pointer, tooltip 35%. MagnetBubbleTests 66/66, project 7180 green, 0 build errors. Known, left: on vertically stacked monitors the cursor test uses the motion-exit bounds (top slack ~250 DIP), a narrow band where a magnet can chase a cursor on the screen above. Desk run owed.

PR #1280, opened 2026-09-15
feat(backroom): slot on a phone - whole-cabinet tap, pan-in instead of a card, sideways nudge and landscape frame

CodeBambi:
## What

Three things from the owner's phone desk run at the slot cabinets (Candy Rose / Violet / Mint), portrait Safari.

### 1. The whole cabinet is the tap target (root cause)

Tapping a cabinet only entered it near the base. The room's bulb auras are one scene-level `Points` cloud (`fixtures.js createAuras`), and a three.js Points raycast answers within `Raycaster.params.Points.threshold` (1 m) of every aura, so from the approach pose nearly every tap on the cabinet hit glow first, in front of the cabinet and on no station holder. Only the base, more than a metre from any bulb, reached the mesh.

Fix at the root: glow is not a surface, so the aura cloud and the roulette mote field get a no-op `raycast`. A bulb batched into a scene-level InstancedMesh also loses its holder, so the batch records its rows and `scene.js` maps a tap on an instance back to its station. No hitbox is enlarged. Measured at 390x844: a 13x18 pick grid over Candy Rose went from "Points everywhere but one row" to "station holder in every cell".

### 2. No "Preparing the cabinet" card: the pan-in is the load screen

- `stations/slot/station.js` exports `roomBehind = true`; the loader keeps its root see-through (`.br-through`) instead of the ground colour, so the room's pan-in runs while the cabinet loads and the cabinet rises over the room's held frame (one line in `loader.js`, one rule in `room.css`).
- `seat-camera.js`: the slot gets a pan-in target (reel window + jackpot screen) fitted on height alone (`travelOnly`), so a narrow phone still travels in: 1.95 m to 1.26 m from the cabinet, eye 1.65 to 1.35 m. Kept to three scoped lines to stay clear of the roulette lane.
- The room still holds on arrival (one drawing context), Back pans out to the exact walking pose, Motion off and reduced motion snap both ways.
- `br_slot_loading` only existed in en.json; removed with its markup (the lexicon test forbids dead rows).
- Also fixed on the way: the walking pills (Room view / Motion / Options / bell) were showing over held stations since the approach pan landed, because `html.br-seated .br-room-ui` outranked `html.br-visiting`; the visiting rule now wins.

### 3. Sideways nudge + landscape frame

- Portrait on a coarse pointer shows `br_slot_rotate` ("Turn your phone sideways for a bigger view", all nine locales) once a session; the whole pill dismisses it (sessionStorage), landscape hides it, a desk never sees it.
- Landscape on a phone (height <= 500 px): pills, Freeze and Odds stack in a 110 px left column, Spin and the face keep the right corners, and the reel window fills the band between using the seat-camera mechanism (a span per axis on the fit plus `camera.setViewOffset`, `scene.js sideband`). Measured at 844x390: reel window 504x234 px (60% of the height) vs 260x121 in portrait, clear of every control and of Back / SP. The spiral jar stands right of the column.

CONTRACT 10.20 records the amendments.

## Tests

- `node --test` over every `*.test.mjs` under backroom: 304 pass.
- New `smoke/slot-entry-check.mjs`: 150 checks pass (five tap points x three cabinets x two orientations = 30 real touch taps each posting `station-open`, pan-in sampling for card / ground colour / distance, pan-out pose, motion off + reduced, nudge session memory, landscape boxes). Screenshots in `_evidence/slot-entry/` (untracked).
- `smoke/room-stations-check.mjs` 98 ok, `smoke/touch-control-check.mjs` 26 ok, `smoke/seat-camera-check.mjs` (see the last comment on this PR for the result).

## Not touched

The roulette station and its branches in `seat-camera.js` / `loader.js`.

https://claude.ai/code/session_01Xz6gHKA8jvJTENAAqtnbVr


CodeBambi:
smoke/seat-camera-check.mjs rerun on 70866f2c0: exit 0, 78 ok, no page exceptions (the first run died on a port clash before any check, exit 13).

PR #1324, opened 2026-09-16
feat(backroom): the reward pass, and the owner's phone pass on the real build

CodeBambi:
The Back Room reward pass, plus the owner's first desk pass on the real web build.

Thirteen commits: five station lanes and the room lane onto one spine, the integration fix, and today's phone pass on top.

## The reward pass (commits 1-12)

One tier, one plan, one bank, one ladder, shared by every fixture.

- **`shared/win/`** is the spine: `tier.js` grades a pay, `plan.js` turns a tier into a party the house has agreed to pay for (Brake 3's ledger lives here now, one of it, not a pair of counters per station), `bank.js` counts the value across, `ladder.js` is the chime.
- **Every station spends the same vocabulary.** The slot puts the reference cabinet onto the spine and picks up three moves it had never spent. The wheel was holding half the vocabulary already. Roulette and cards pay out loud, with a party the house sizes rather than the table.
- **The whole floor hears a win** (`room/win-echo.js`): a pay anywhere raises the room, Brake 2 merges two pays landing on one fixture inside one beat, and THE REVEAL leaves the fixture once a visit.

## The phone pass (commit 13)

Seven findings, every one a divergence from something already written down rather than a taste call.

- **Reel spirals are Loom fields.** `symbols.js` drew an Archimedean coil by hand; every other spiral in the room comes from the Loom (CONTRACT 10.13.D). One 256 px tile per preset per repaint clock, so a strip of thirteen spirals is one GL render, not thirteen.
- **EMI is framed in on a phone**, in the pose each orientation can afford. Upright she keeps her topper and the topper is measured into the fit box; sideways she takes the shelf. Giving portrait the shelf instead cost the reels a fifth of their size, measured - the shelf is outboard and portrait has no width to sell.
- **The spiral jar tube is gone.** On a phone it stood exactly where EMI's shelf is. The jar is untouched; the count reads as its own chip, which is all Brake 9 ever asked of it.
- **Wall screens move and turn.** `TURN_S` was 18 s and the decode budget was a second throttle stacked on each source's own clock, so four visible pictures ran at three frames a second each.
- **The chrome has a body**: one depth recipe across the room shell and all five stations.
- Chase bulbs 1.5x in the scene, not the glb. Landscape side column 110 -> 152 px in all three places that hold the number.

## Tests

563/563 unit tests green (`node --test` over every `*.test.mjs` under `backroom/`).

Verified live on a headless phone at 390x844 and 844x390 against the deployed playtest build: page interactive, zero page errors, all five stations answering the real engines, wall pictures animating at 9.6 frames/s facing a screen bank on a 4.6 fps software-GL loop.

## Known and deliberate, not fixed here

Naming these so they are not discovered as surprises in review:

1. **The slot wears Brake 3 twice when the jar and a line land together.** The jar's tier 2 party borrows `spiral3`'s shape because it *is* that event (10.16.A), and both it and the line settle against the one ledger.
2. **`room/win-echo.js` over-counts on a merged echo.** `afterParty` runs on the merged plan, but the first pay already spent its own rung when it started, so one moment wears the ledger twice.

Both are ledger arithmetic, not anything the player sees misfire, and both want a decision about which event owns the rung rather than a patch.

## Owner decisions still open

- Does Calm mean lite everywhere, or is lite its own level?
- Does Calm silence the slot jackpot pop?
- Brake 9's room-side text channel is dead code right now - keep it or cut it?

## Not in this PR

- `SPIRAL_STRETCH = 1.8` in the web host stand-in is a deliberate divergence from CONTRACT section 4's authored 4 s / 1.5 s, marked as one in its source. If the longer hold is right, section 4 and the WPF host are where it has to be written down.
- The desk run of the reward pass in the actual WPF app is still owed. Everything above was judged in a browser.

https://claude.ai/code/session_01ScWoVAjo9r576u8hCJc5JY


CodeBambi:

PR #1329, opened 2026-09-16
feat(backroom): play the room slot in a compact closeup

CodeBambi:
Entering a slot now pans and zooms onto that same room cabinet and plays its existing reels. The approved closeup fills the viewport with larger reel windows, smaller EMI on the lower-left ledge, three compact Freeze controls and one Spin button. Back, Jackpot, Spirals and Last win sit across the top; utilities sit bottom-right.

The cabinet adapts to portrait and landscape screens using the room renderer. Leaving restores the room camera, fixture proportions and visibility while retaining the final reel image and stop. No additional WebGL context or replacement play cabinet is created. Tape results, odds and payouts are unchanged.

Validation: 587 Back Room tests passed; all three cabinet colours checked across desktop and both phone orientations, with focused final portrait and rotation checks. Browser checks covered input, exact reel identity and stop retention, Motion Off, room return and accessible controls. Fresh GLB import passed with 41,514 triangles. Live preview checks passed at three viewport sizes.

Known visual limitation: the crown trim shares merged cabinet geometry, so its shape stretches with the portrait layout.

Stack 2/3, based on #1328. 404 changed text lines plus the cabinet GLB. Browser regression harnesses follow in the next PR. Separate prize-unlock work is excluded.

Preview: https://backroom-playtest.vercel.app/backroom/index.html

PR #1330, opened 2026-09-16
test(backroom): cover shared slot gameplay and responsive layout

CodeBambi:
Adds browser regression coverage for the shared slot cabinet introduced in #1329. Checks exercise three cabinet colours, desktop and both phone orientations, lever and Freeze input, reel identity and animation, final strip retention, Motion Off, camera restoration and unobscured controls.

A separate real-host layout check visits and leaves the slot without spinning. It accepts SHARED_SLOT_URL or SLOT_LAYOUT_ROOT, uses an isolated browser profile, and closes only its own child process.

Validation: full nine-case mock-host run passed, followed by focused final portrait and rotation checks; real preview passed at three sizes with no page errors. The real-host harness received a syntax check after removing its unused mock setup and requiring an explicit preview location.

Stack 3/3, based on #1329. 330 added lines. No runtime changes.


CodeBambi:

PR #1333, opened 2026-09-16
feat(backroom): soft Circe preset recordings with fades

CodeBambi:
Replace the missing Back Room preset word recordings with nine soft Circe clips: DROP, RELAX, LET GO, SINK, DEEPER, EMPTY, OBEY, SOFTER and BLANK. The manifest feeds the existing desktop clip resolver, preserving player-owned clip priority.

Each recording has a 55 ms fade-in and 140 ms fade-out, a quiet normalized level and restrained high frequencies. Files are tracked through Git LFS; no keys or paid generation scripts are included.

Stacked on #1332 (table feedback). The later blackjack/preview integration PR preserves the browser playback helper and its bridge wiring, so both desktop assets and the deployed web playback are reviewable.

Validation: all nine manifest files exist; durations and levels were checked with ffprobe/ffmpeg. Browser playback/reversal/cancellation has four passing tests in the final stack. Human delivery preference remains an ear check; no WPF application was launched.

PR #1334, opened 2026-09-16
feat(backroom): blackjack table controls, handling and look-around

CodeBambi:
Blackjack now shows the active hand total on the felt, a 1-3 SP +/- wager with one chip per SP, and prominent Hit/Stand controls with Double/Split underneath. Chips reserve space outside six-card and split-hand footprints. Their add/remove, Double, split and settlement movements follow the actual wager and per-hand server result. Dealer reveal contact and active-hand feedback finish the physical pass.

Both card and roulette tables allow a small eased look-around by dragging empty felt, with Center view. Game targets keep their gestures. Motion Off/Calm, suspension and camera transitions reset the look; leaving cleans up capture and listeners.

Also fixes the table freeze during delayed state loading: the animation loop tolerates the bet output not existing yet. Browser checks delay state replies by 650 ms to cover that race. The preview-only Circe helper and documented bridge hooks are preserved under scripts/backroom-playtest, with playback, cancellation, reversal and missing-clip tests. Private server and ledger implementation stays out of this public repo.

Stacked on #1333 (Circe recordings). Requires the private server three-chip PR for account-backed 3 SP play; the client honors advertised stakes and remains capped at 2 until that server change is deployed.

Validation: 608 Back Room unit tests plus 4 preview voice tests pass. Desktop and phone blackjack gameplay checks cover hit, split, double, reveal, Motion Off and immediate Back. Six-card chip-clearance assertions pass. Focused roulette look/recenter/Back checks prove that looking places no wager. The combined latest-preview check passes and 11 deployed files match their expected hashes. No WPF application was launched.

The combined playtest is published at https://backroom-playtest.vercel.app/backroom/index.html?juice=pass2. Git/PR creation does not deploy the client or private server. Other task slot/media refinements are preserved in the preview and remain in their own source changes.


CodeBambi:

PR #1337, opened 2026-09-16
Add pedestal decorations and fix Remove/Use visibility

CodeBambi:
Expose the three existing statue-and-pedestal assemblies on the Decorations shelf. Remove now hides the entire assembly, and Use restores it. Stop owned-prop previews from immediately making removed decorations visible again. Add the four control/item labels in all nine languages. Ownership and reward rules stay unchanged.

Validation: Browser checks pass for all six collectible props and three pedestals through Remove and Use, including visibility while the item remains selected. All nine language files parse.

Stacked on `codex/backroom-slot-polish`. 62 changed text lines relative to this PR base; binary model changes are separate from that count. Merge bottom-up after review.

Client source only. Private preview-host/server files are excluded. No deployment or merge is part of this PR.


CodeBambi:

PR #1340, opened 2026-09-17
The casino batch: a slip at the blackjack table, spins earned at the tables, a reachable shelf, softer spirals, distinct burst pictures

CodeBambi:
Five items from the 2026-09-17 batch, client half. The server half is CC-Labs-llc/CCP-Server `feat/casino-batch-0917`, and the two land together: the shelf prices and the free-spin wire are agreed across both.

**607 changed lines, 7 over the 600 cap.** It is five features in one branch because that is what was asked for; if you would rather review it in pieces the natural split is (5) and (4) as one C#-only PR, (3) and (2) as the economy PR, and (1) on its own.

---

### 1. THE SLIP: a blackjack Easter egg

About one settled hand in fifty, one card on the felt tears and comes back as a different card. New pure module `stations/cards/glitch.js`; both renderers read it, so the 2D felt and the 3D faces agree.

The GIF comes free. Both renderers already resolve a card's picture from its code (`deck.draw(g, deck.keyFor(code), ...)`), so swapping what the card wears swaps the picture with it.

Cosmetic by decision and by construction. The slip is applied at **paint** time and never written back to a code, so the felt total, the legal moves and the pay all still read what the server settled. A client that could really change a card would be a client that could cheat.

One consequence worth knowing: because the total is drawn from the true codes, a slip that changed a card's *value* would print a face that disagrees with the total under it. So the draw prefers a same-value replacement that still moves the picture, which is the ten family (T J Q K share a blackjack value but not a deck slot), and only falls back to a free rank when that family cannot serve. `planSlip(..., { free: false })` forbids the fallback outright if you would rather the felt be provably honest.

Armed on `settle`, after the decision closes, so a tear can never sit beside a live choice.

### 2. Free wheel spins earned at the slot and the roulette

CONTRACT 10.16.H. The wheel stays daily; the tables bank another spin. Both triggers ride the one authored moment each station already has, so the grant always lands with a celebration playing.

| Station | Grant | Rate |
|---|---|---|
| slot | a `spiral3` on a spin that was **paid for** | 8,479 over DEN, **1 in 118** |
| roulette | a **woken straight**, the only outcome that fires `fx.jackpot` | about **1 in 765** spins carrying a straight bet |

Capped at 3. A plain Spiral Wake is 1 in 20.7 and far too rich for a daily wheel, which is why the straight is what makes the roulette grant rare. Expansion, free, jar and comp slot spins grant nothing: they cost 0 SP, so a `spiral3` inside one would mint wheel spins out of nothing.

Page side, the station now asks one question, `spent()`: today's spin is gone **and** nothing is banked. While a bank is held, a spun wheel reads exactly as an unspun one.

> This reverses one line of `WHEEL-PRIZES.md` ("No extra spins or guest passes"). Flagged because that doc is otherwise the spec for the branch this folds into.

### 3. The shelf a non-casino player can reach

`flashes_v2` and `bubbles_v2` drop **240 → 30 SP** (whole shelf 13,935). At 30, one 30/60/150 wedge covers a power-up outright, the 28.5 SP daily cash EV is about one power-up a day, and the 60 SP average balance buys both on day one. Prices are config now, not literals: see the server PR for `BACKROOM_COUNTER_PRICES`.

**THE NUDGE.** After the first power-up, a balance short of the other one shows one quiet line pointing at the wheel. No button, nothing posted, no station change.

> The arithmetic you asked for is strict, and it has a consequence I flagged while planning: at 30 a row and 60 SP in hand, buying one leaves exactly 30, which is **not below** 30, so the median player never sees this line. It only fires below a 60 SP starting balance. There is a test pinning that behaviour as deliberate. Widen it by moving the price in `BACKROOM_COUNTER_PRICES`, not by loosening the comparison.

### 4. Spirals fade further and hold longer

The fade already existed and was correct. The reason it read as a pop is that 250 ms at the head of a 4 s hold is close to instant.

`SpiralFadeInMs` 250 → **600**, `SpiralFadeOutMs` 500 → **900**, and the holds grew with them (`SpiralFullMs` 4000 → **5000**, `SpiralBriefMs` 1500 → **2000**) because the rise is spent *inside* the hold. Total on screen is hold + 900.

The jackpot hero's own spiral is deliberately unmoved: it runs for `JackpotHeroMs`, so its burst, rain, glitches and words still land together on its beat. There is a test pinning that.

### 5. The roulette burst shows five different pictures

Root cause, and it is not bad luck. The burst carries no picture at all (`BackRoomFx.cs:239` hands `FlashBurst` a count and no gif), so it lands on the user's own flash pool. That pool dedupes on **path string**, and `ContentPackService.GetPackFileTempPath` mints a fresh `ccp_temp_<guid>` path on every decrypt, so the same pack picture drawn twice looked like two different files and walked straight past the dedup. Pack and curated users saw repeats; a plain-folder user with five images never did.

An authored burst now draws from `GetChaosImagePaths`, which already dedupes on **source identity** (pool index for disk and pack entries, URL for remote) and exists for exactly this reason. Ambient flashes keep their with-replacement picks, which is deliberate.

One accepted trade: with a pool smaller than five the burst now shows fewer images rather than repeating.

---

### Tests

- **7298 C# tests pass**, 0 fail, 1 skipped.
- **442 backroom web tests pass**, 0 fail.
- New: `stations/cards/tests/glitch.test.mjs` (9), the wheel bank in `stations/wheel/tests/mock.test.mjs` (2), THE NUDGE in `stations/counter/tests/cards.test.mjs` (3).
- Counter fixtures were rebased on the new price rather than having their expectations loosened: the "short by N" cases still exercise a short balance.

### Not done

- `br_counter_wheel_nudge`, `br_wheel_banked` and `br_wheel_banked_n` are English in all nine language files. Translations owed.
- `stations/cards/table.js` is also being edited on `codex/backroom-juice`. My hook there is four lines, but that branch should merge first or be merged into this one before either lands.

https://claude.ai/code/session_01AiLm2SY5fwLTiTYRhPXq2Q


CodeBambi:
Folded into integrate/race-stack-0918 at 88eeb4b3a (tracked as #1381), Astra's side on every collision: slip, distinct burst pictures and the non-casino shelf are in; spiral envelope, banked free spins and the wheel nudge are out. Server half #191 is still its own decision.

PR #1346, opened 2026-09-17
Celebrate Back Room prize purchases and shelf arrivals

CodeBambi:
Prize purchases now emit a short diamond burst. Racing Thoughts demo and expansion models travel from the counter to their owned shelf, alongside the existing lift, turn, SOLD stamp and cabinet reveal.

Motion Off settles immediately. State refreshes do not replay the celebration, and all owned particle resources are disposed with the display.

Validation: prize display lifecycle tests cover flight completion, Motion Off, state refresh, particle expiry and cleanup; 4/4 pass.


CodeBambi:

PR #1343, opened 2026-09-17
Back Room: sparkle wallet, wheel effects and arcade placement

CodeBambi:
Move the prominent sparkle wallet and controls above the prize sheet, show signed transfers, and give the wheel animated slice highlights and diamond trails. Bonus credits enable wheel spins during the daily wait.

Keep the locked arcade behind the counter, then reveal the cabinet and owned-model shelf against the wall after purchase.

Validation: 635 Back Room tests and browser checks at desktop/phone sizes passed, including bonus earning and spending. Draft, no deployment.

PR #1341, opened 2026-09-17
Back Room: blackjack controls and prize previews

CodeBambi:
Make blackjack totals and side actions clearer, slide the previous hand away before dealing, and show moving V2 prize previews with readable descriptions.

Includes the translated copy for the following slot and wheel changes in this stack.

Validation: integrated Back Room suite 635 passing; desktop and phone preview checks passed. Draft, no deployment.

PR #1339, opened 2026-09-16
Animate prize purchases and reveal the owned arcade display

CodeBambi:
Confirmed purchases animate once and mark the corresponding shelf model SOLD. Demo ownership removes the demo listing/model and reveals the arcade placeholder beside the counter with a fade and rise; the adjacent shelf displays owned expansion models. Existing ownership restores silently. The cabinet is a display only: Racing Thoughts and its levels remain in CCP Play. Preserve the newer counter delivery feedback while adding the ownership reveal.

Validation: All 622 Back Room unit tests pass on this integrated head. Prize browser checks pass for purchase, bundle-provided demo ownership, reveal, owned expansion display, reachable approaches, cabinet clearance, phone layout, Motion Off and zero browser errors.

Stacked on `codex/backroom-polish-checks`. 444 changed text lines relative to this PR base; binary model changes are separate from that count. Merge bottom-up after review.

Client source only. Private preview-host/server files are excluded. No deployment or merge is part of this PR.

PR #1365, opened 2026-09-17
Add the Racing Thoughts cabinet and prize display support

CodeBambi:
Bring the minimum cabinet and owned-prize display support into this main-based racing stack. Register the racing cabinet as a reachable scene target and expose a room pose seam for the following bridge.

Validation: Prize state/display tests and the integrated local browser interaction checks passed. The newer combined preview has existing camera-assumption failures in its older room smoke; those are not claimed green.

Stack slice 8/10; base `codex/race-07-desktop-access`. 242 changed text lines, plus the counter GLB asset. Review and land bottom-up; retarget each child after its parent lands. Draft only; nothing merged or deployed.


CodeBambi:

PR #1366, opened 2026-09-17
Open Racing Thoughts from the casino in the same window

CodeBambi:
Clicking the owned racing cabinet transfers the existing desktop window into Racing Thoughts, then returns to the casino after normal race cleanup. Canonical grants authorize entry; cursor flushes, effects, banking, pending callbacks and return pose follow their existing lifecycle boundaries.

Validation: Integrated native bridge/access suite passed 39/39 and payout regressions passed 2/2. Combined preview round-trip checks passed desktop 27/27 and phone 30/30. Actual WPF window transfer still requires the owner desk check.

Stack slice 9/10; base `codex/race-08-cabinet-support`. 437 changed text lines. Review and land bottom-up; retarget each child after its parent lands. Draft only; nothing merged or deployed.

Final native boundary verification: this exact commit built independently in a clean detached checkout, Release 0 errors and 362 existing warnings.


CodeBambi:

PR #1367, opened 2026-09-17
Save the local casino racing preview and round-trip regression

CodeBambi:
Save the reproducible local casino racing adapter, staging script and precise integration patch, plus the isolated desktop/phone round-trip harness. Generated racing assets remain outside Git; the adapter uses only the existing local preview ledger and never awards account currency.

Validation: Syntax checks passed. Staging into an isolated scratch preview produced 272 files / 27,340,033 bytes; generated output was not committed. The saved harness previously passed desktop 27/27 and phone 30/30 against the combined preview.

Stack slice 10/10; base `codex/race-09-casino-bridge`. 373 changed text lines. Review and land bottom-up; retarget each child after its parent lands. Draft only; nothing merged or deployed.


CodeBambi:

PR #1368, opened 2026-09-17
Three bug fixes for the casino pre-release

CodeBambi:
Three fixes from the ccp-bugs queue, folded in ahead of the casino pre-release. Independent, one commit each.

### `fix(programs)` the mantra task card opens the mantra game (#1230)

Day 5 of Kept says "Complete 3 mantras" and the how-to under it says to type them in the mantra minigame. There was no mantra minigame to open. The Play page's Mantras card was `MantraWindow`'s only launcher and it came off in the 2026-08-12 relayout, leaving the window with no caller while Kept days 5, 15 and 27 kept asking for it. Only the voice path credited anything.

The rescue note in `MainWindow.PlayTab.cs` said re-homing the game cost exactly one `StartMantraSession(reps)` call. This is that call, on the card that asked for the mantras and knows the rep count. Shown only for an incomplete, unblocked Mantra task on an unpaused run; the voice path is untouched.

### `fix(deeper)` a rule-fired speak set to "Pause the video" actually pauses it (#1221)

A speak on a timeline band carries a region, so the hold watch had a boundary to police. A speak fired by a rule does not - `EnhancementEngine` hands it a null region end when the item has no authored width, and `StartHoldWatch` returned on exactly that. Cue up, listener open, video playing straight on underneath.

There is no boundary to wait for in that case, so the hold is taken at cue time instead. The existing 30s hard deadline still bounds it and every exit path already runs `ReleaseHold`. `LoopRegion` still needs bounds and stays a no-op without them, as before.

### `fix(deeper)` the wheel scrolls the page over the library list (#1222)

The library list templates its own `ScrollViewer` inside the tab's page `ScrollViewer`, so it is measured unbounded and its inner viewer has `ScrollableHeight` 0 - and WPF marks a wheel notch `Handled` even with nothing to scroll. Every notch over the rows was swallowed. Reuses `CompanionWheelRelay`, which was extracted for this exact shape.

### Verification

- Build: 0 errors.
- Full unit suite: **7300 passed, 0 failed, 1 skipped**.
- Two new render tests cover the mantra door: one asserts a Mantra card carries a visible button bound to the task's reps, one asserts an ordinary card carries none. Both realize the real `ItemTemplate`, so deleting, collapsing or unbinding the button fails them.

Desk verification still owed for #1221 and #1222, which are pointer and playback behaviours the suite cannot synthesise.

### Deliberately not in this PR

**#1223** (fullscreen toggle puts the video over the overlay effects) is the fourth report of #1041/#1051/#1052. The reassert machinery already exists in `EnhancementPlayerWindow`; this is a timing race and wants a desk repro, not a blind patch days before a showcase.

**#1216** ("haptics dont work", v6.9.3) has no steps and no repro.

https://claude.ai/code/session_01Xz6gHKA8jvJTENAAqtnbVr


CodeBambi:

PR #1395, opened 2026-09-18
fix(backroom): wall pictures play at their own rate, up to 30 fps

CodeBambi:
Desktop tester feedback: "general framerate is fine, framerate on the images on the various monitors is awful. especially noticeable on the spiral gifs".

## Root cause
Two things in `room/gif-decode.js`, both per wall source:
1. `MAX_FPS = 12`. A spiral gif is authored at 25-33 fps (30-40 ms delays), so the cap alone made it lurch.
2. The next frame was scheduled from decode **completion** (`dueAt = performance.now() + delay`), not from the frame's own due time. Decode latency plus the 30 Hz rest render tick (the source only advances when the room renders) compounded, so a "12 fps" source really ran at 8-9 uneven fps.

Also checked and cleared: `quality.performance` is not accidentally on for desktop by default (it is `mode === 'performance'` or auto + reduced; reduced starts from `createRenderBudget().mobile`, see caveat). Off-frustum screens still cost nothing (frustum check unchanged). Texture upload is a same-size `CanvasTexture` at 384 px max, which three already re-uploads with `texSubImage2D` on WebGL2, so no ImageBitmap cache is needed.

## Fix
- `MAX_FPS` 12 -> 30 for the room's wall sources. The slot (`maxFps: 12`) and hypno deck (`DECK_CAPS.maxFps: 12`) pass their own cap and are unchanged. Clips (`clip-source.js`) follow the room cap.
- Next due time is anchored to the previous due time (`dueAt = (late > step ? now : dueAt) + step`), so a source plays at the file's own frame delays up to the cap with no latency drift; when more than one delay late (off-frustum, held room) it re-anchors to the waking tick rather than bursting to catch up.
- `screen-animation-budget.js`: Full allows eight starts per rendered frame (every dealt picture, `MAX_PICTURES`), so the per-frame batch is no longer a second throttle at a 30 Hz rest render. Performance keeps four starts per 1/6 s, so the shared clock still bounds total work on a slow device.
- `clip-source.test.mjs` timing updated for the new gap; CONTRACT.md 10.x GIF caps and room/PERFORMANCE.md amended.

## Tests
`node --test` over every `*.test.mjs` plus `room/lever-return.test.js` from the backroom dir: **832 pass, 0 fail** (previous known green count 824; main has grown).

Skipped `smoke/perf-bench.mjs`: it runs against the smoke checks' fake host, which deals no animated pictures, so it would not measure this change.

## Caveats
- Upper bound of decode work in Full rises from 4 starts/frame to 8 (only if eight 30 fps pictures are all in the frustum at once); each is still one composited 384 px frame.
- `createRenderBudget().mobile` flags any machine with `hardwareConcurrency <= 4` or `deviceMemory <= 4` as conservative, and in Auto that machine starts in Performance and never recovers. A 4-core desktop therefore gets the phone budget (four starts per 1/6 s) unless the player picks Full in the options card. Not changed here; worth a separate look if a tester on such a box still reports choppy pictures.


CodeBambi:
Reviewed for the 6.10.1 pre-release. 27 lines: room wall pictures were capped at 12 fps by gif-decode.js MAX_FPS and rescheduled from decode completion; now 30 fps for room screens with due-time anchoring. CI green. Merging.

PR #1396, opened 2026-09-18
feat(backroom): slot "Keep playing", no-stake spins that run the effects

CodeBambi:
## Desk tester feedback

> the ccp effects on the slots are good, but knowing you have limited spins limits their effectiveness - can't sink into it when you know it'll be over before you properly fall

## The mode

**Keep playing.** After the first landed row a small `Keep playing` key sits over Spin; while it is on (`Play for SP` turns it off) every press deals a no-stake row. The same happens on its own when the tape is empty and no spin can be bought (no comp, no free spin, not one SP). The reels, the anticipation hold, the glyph hit, the callout, the spiral / tunnel / gif burst / word fx and the landing hold all run exactly as for a paid row, because the row goes through the same `beat()` / `flow()` / `land()` and the same client-side plan builder (`feel.flowPlan`).

What a demo row does NOT do: buy a tape, register a chase, fly THE BANK, rain the room's coins, tick the jar, or ask the host for a re-deal. The key's label and a cabinet line read **"Free play, nothing to win"** before the press (`br_slot_freeplay`, all nine locales), and freezes are off in free play (a freeze is a server buy).

**The ending is not advertised while it plays.** During a landing hold (`feel.unlockMs` from the landing) the "N prepaid spins left" line under Spin is blank and the status / state note / comp chip soften (`data-hold`); they come back on the idle screen. `flow()` schedules one `sync` tick at the unlock so they return without waiting for input.

## Money safety

- `stations/slot/tape.js` is untouched. The demo path (`stations/slot/freeplay.js`) stands beside it: it has no `request`, so no `slot/spin`, no `slot/chase`, no ledger op can come from it. `wheel-chase.test.mjs` still guards the chase refusal.
- Every demo row is `pay: 0`, `kind: 'demo'`, so `feel.tierOf` reads 0 (no tokens, no bank, no shower); `land()` carries a `demo` guard on top of that anyway. The row is not one of the tape's own objects, so `tape.land()` refuses it and the cursor, melt, jar and free spins stay where the server left them. `emi3`, `emi2` and `melt` are never dealt.
- The press predicate and the key's label read the same snapshot, so what the key says is what the press does.

## Tests

`stations/slot/tests/freeplay.test.mjs` (4 tests): forty demo spins leave the request log untouched, the tape refuses each row and its snapshot is byte-for-byte the same; every demo line deals off the strips, reads as its line under `mock-server.lineFor`, and yields an effects plan; the dealer refuses rather than invents a row; the counter gate hides during spin / reveal / an unfinished hold and returns on idle.

Whole web unit suite from the backroom dir: **832 / 832 pass** (107 files, `node --test`).

## Caveats

- No browser smoke run (per the lane rule). The keep key and the free-play line are positioned for desk, phone portrait, short landscape and the shared-room projection in `station.css`; worth one look in the room on a phone.
- A demo row with a paying line (e.g. spiral3) shows the line's callout and fx but the cabinet's loss recipe (muted thud, no coins), which is the honest reading of "nothing to win".
- While `Keep playing` is on, a prepaid tape or free spins still owed simply wait; `Play for SP` resumes them.


CodeBambi:
Reviewed for the 6.10.1 pre-release. Slot "Keep playing": no-stake demo spins in stations/slot/freeplay.js, same fx plan, pay 0, no server op, tape.js untouched, spins-left line blank during a hold. CI green. Merging.

PR #1399, opened 2026-09-18
fix(backroom): desk station Backs, a loss callout at the card table, chip never stuck at 0 after a win

CodeBambi:
Desktop tester feedback on the Back Room (2026-09-18). Builds on #1386 (merged into this branch).

## (a) "slots have a Back button but wheel of prizes doesn't"
Every seated station (wheel, cards, roulette) and the counter now keep their own Back in the room's top-left corner, the way the slot already did (`root.dataset.ownBack`, pinned by the #1386 rule in `shared/back.css`). `shared/back.css` hides the room's `#br-back` under a station with `data-own-back`, so there is exactly one Back per station and the slot is not doubled (it already hid `#br-back` in its own css).

## (b) "card game gives a Winner notification but not when you lose"
`stations/cards/feel.js`: `cards.lose` -> `br_callout_dealer_wins` ("Dealer Wins") and `cards.push` -> `br_callout_push` ("Push"), both `tier: 'small'` like Winner, fired on the settle frame next to the loss thud. New keys added to all nine locale files with the English text. The party/celebration text stays win-only.

## (c) "winning the card game seems to reset my SP to 0 until I stand up; I can still bet"
Root cause: on the settle frame the station pins the room's chip to `before` = server balance minus what the hand still owes (Law I), which is exactly 0 when the stake was the whole balance, and the only thing that hands the number back is THE BANK's `done` (`onDone -> chip.show(null)`). Any path where the flight value outlives the bank left the chip on that pinned 0 until `close()` -> `chip.handOver()` (standing up), while the bet gate and the server still knew the real balance. Fix: the station's frame loop returns the chip to the rule (server minus owed) whenever a flight value is set and no bank run is busy, so the shown SP is the real balance and `chip.value` (the bet gate) is the same number the room shows.

## Tests
- New `stations/cards/tests/chip-handback.test.mjs` (every bank exit hands the settled value back; the rule is never below zero and equals the server once nothing is owed).
- `feel.test.mjs` covers the lose/push callouts; `tier`, `presentation-pacer` and `counter/station` tests updated for the new behaviour.
- Whole web unit suite: 833/833 green.


CodeBambi:
Reviewed for the 6.10.1 pre-release. Merges clean against main (verified with git merge-tree; the DIRTY flag was stale). CI windows green. Station-owned Back on desktop via data-own-back, cards loss/push callouts, and the defensive chip hand-back. Merging as the base of the desk-feedback train.

PR #1394, opened 2026-09-18
feat(deeper): player keyboard shortcuts and hold-Esc

CodeBambi:
Part 3 of 3 of the Deeper player wave 1 pass. Base is `feat/deeper-player-wave1-qol`. ~370 changed lines.

## What changed
- **Keyboard shortcuts** on the player window: Space play/pause, Left/Right seek 5 s (Shift 30 s), Up/Down volume 5, M mute (restores the previous level; a muted 0 is never persisted), F fullscreen toggle (video), Esc leaves fullscreen or closes the window. Keys are ignored while a TextBox/PasswordBox has focus.
- **Page-side relay**: WebView2 owns focus in video mode, so the injected script posts `ccp_key:<name>:<shift>:<repeat>` / `ccp_keyup:Escape` over the existing postMessage bridge and C# handles the same actions. F tries HTML5 fullscreen in-page first (user gesture), falling back to the forced reparent.
- **Hold-Esc escape hatch** (community ask, people get stranded in blink/gaze loops): holding Esc for 3 s anywhere in the player (window, fullscreen host, or page) stops playback and unbinds the engine, with a "Hold Esc to stop" progress hint in the info strip. Off while `App.Lockdown.IsActive` or `StrictLockEnabled`. A lost key-up (focus swap mid-hold) is treated as a release after 1.5 s of silence.
- The fullscreen host gets the same bindings (F11 still exits).
- Shortcut list is appended to the play button tooltip (`deeper_player_tip_shortcuts`).

## Tests
- `dotnet build` clean; full unit suite: 7521 passed, 0 failed, 1 pre-existing skip.
- Keyboard/JS relay paths are UI-only and not unit-tested; not exercised in the running app.

PR #1400, opened 2026-09-18
feat(deeper): library wave 1 polish

CodeBambi:
Part 1 of 3 of the Deeper library wave 1 pass (behaviour). Stack: this PR -> #1401 (row interaction) -> #1402 (filters/sort/empty state). Independent of the player stack (#1392-#1394). ~630 changed lines, of which ~190 are the new pure helper + its tests; the app-side delta is under the cap.

## What changed
- **Import feedback** uses `App.Notifications` toasts like the rest of the tab (no more stock `MessageBox`); the imported row is scrolled to and briefly flashed (`DeeperLibraryRowVm.IsHighlighted`, `RevealDeeperLibraryRow`).
- **Non-enhancement JSON is skipped quietly**: `EnhancementImportRules.FileLooksLikeEnhancement` sniffs the first 4 KB for the `$schema` tag (same check `EnhancementSerializer.Load` makes), so a dropped `settings.json` gets an Info toast instead of an "Import failed" modal.
- **Re-import dedupe**: `EnhancementLibrary.FindDuplicateOf` returns the existing library path when the source already lives in the library folder or a library file has identical content (whitespace-insensitive SHA-256). The import then toasts "already in your library" and reveals that row instead of writing "Name (2)".
- **Delete goes to the Recycle Bin with Undo**: confirm via `ShowStyledDialog` (Delete / Cancel), the row hides immediately, the file is moved with `Microsoft.VisualBasic.FileIO.FileSystem.DeleteFile(..., SendToRecycleBin)` after a 6 s grace, and the toast's Undo cancels it. Pending paths are skipped by `ReloadDeeperLibraryFromDisk` so a watcher refresh cannot resurrect the row. The orphaned `DeeperSubmissions` record keyed by the path is removed. Closing the app inside the 6 s window leaves the file in place (safe direction).
- Demo-missing message goes through `ShowStyledDialog` and drops the em-dash; import dialog title and every import string are localised (13 new keys in all 9 language files, English fallback).
- **Log hygiene**: `EnhancementFetcher` no longer logs the enhancement name; `EnhancementLibrary`, the tab and hub drop absolute paths from log lines; empty catches in Tab/Hub/Submissions -> `Diag.Swallowed` / typed catches.
- `en.json`: `deeper_library_delete_confirm_fmt` no longer claims the delete is permanent (other languages keep their text; they now overstate the risk, which is the safe direction).

## Tests
- New `DeeperLibraryWave1Tests`: extension gate, schema sniff (accepts the tag, rejects settings-style JSON / json-schema `$schema` / empty), normalized hash ignores formatting and differs on content, inside-folder check, on-disk sniff.
- `dotnet build` clean; full unit suite: 7529 passed, 0 failed, 1 pre-existing skip.
- Not exercised in the running app (no UI smoke in this pass).


CodeBambi:
Superseded by #1406

PR #1401, opened 2026-09-18
feat(deeper): library rows - selection, keyboard, context menu

CodeBambi:
Part 2 of 3 of the Deeper library wave 1 pass. Base is `feat/deeper-library-wave1` (#1400); merge that first. ~410 changed lines.

## What changed
- **Selection model**: single click selects a row (accent border + tint via `DeeperLibraryRowVm.IsSelected`), double-click opens the editor. Selection is kept by path across filter/sort rebuilds.
- **Keyboard**: with the list focused, Enter opens, Delete deletes (same confirm as the button), Up/Down move the selection and scroll it into view; Ctrl+F anywhere on the tab focuses the search box.
- **Right-click context menu** on rows, dark-styled (WPF's stock menu is white): Open in editor, Play, Reveal in folder (Explorer `/select`), Copy path, Submit to catalogue (visible/enabled on the same rules as the button), Delete.
- **Hover-only action buttons** are `IsHitTestVisible=False` and `Focusable=False` while faded out, so an invisible button can no longer be clicked or tabbed into.
- **Reuse open windows**: `OpenDeeperEditor` keeps a path -> window dictionary and activates an existing editor for the same file; Play activates an existing player window when `App.DeeperHost.LoadedFilePath` already is that file (no dependency on the player stack's `ShowOrActivate`).
- **Tooltips**: long names get a name tooltip; the missing-media warning glyph explains what it means.
- `en.json`: `deeper_hub_tip_row_open` now describes select / double-click / right-click. 7 new keys in all 9 language files.

## Tests
- No new pure logic to unit-test here (all handlers are UI-bound); `dotnet build` clean; full unit suite: 7529 passed, 0 failed, 1 pre-existing skip.
- Not exercised in the running app.


CodeBambi:
Superseded by #1406

PR #1402, opened 2026-09-18
feat(deeper): library filters, Duration sort, empty state

CodeBambi:
Part 3 of 3 of the Deeper library wave 1 pass. Base is `feat/deeper-library-wave1-rows` (#1401). ~445 changed lines, about half of it the new pure helper and its tests.

## What changed
- **Filter/sort/count logic extracted** into `Services/Deeper/EnhancementLibraryFilter` (pure, unit-tested); `MainWindow.DeeperHub` keeps only UI state and wiring (its two public enums became aliases of the helper's).
- **Pill counts agree with the list**: every pill now counts "rows that would show with THAT pill on and every other active filter kept", so the type pills and the Haptics/Webcam pills can no longer disagree with the rows under them (they used to count against search only).
- **Header count** shows the total, plus "(n shown)" while a filter hides some.
- **Sort**: Duration added; a direction toggle (▼/▲) next to the dropdown flips the order. Recent and Duration default to descending, Name and Creator to ascending; ties fall back to name so unknown durations keep a stable order.
- **Empty state** gets inline New / Import / Browse catalogue buttons, shown only while the library has no files (the "nothing matches your filters" state keeps just the hint).
- **Localised** the remaining hard-coded English on the tab: Import button + tooltip, the webcam card intro, the tracker button's XAML default (note: `MainWindow.BlinkTrainer.cs:RefreshBlinkTrainerTrackerButton` still writes the English "Start/Stop tracker" label at runtime; it is outside this PR's file set and left alone). 7 new keys in all 9 language files.

## Tests
- New `EnhancementLibraryFilterTests`: search over name/creator/tags, pill counts agree with the filtered list under a type filter, a tag filter and a search, Duration/Recent/Name/Creator sort in both directions, stable tie-break, default directions.
- `dotnet build` clean; full unit suite: 7540 passed, 0 failed, 1 pre-existing skip.
- Not exercised in the running app.


CodeBambi:
Superseded by #1406

PR #1425, opened 2026-09-19
Add clip playback and bubble delivery rendering primitives

CodeBambi:
Adds the rendering and pool primitives for animated picture bubbles and Scrolller clips: bounded silent VLC frame playback, consumable remote selections with stale-download rejection, bubble-to-flash geometry, proportional corners, and compositor frame replacement. Includes the Melt sprite and prioritizes less-served flash channels.

This is the foundation of a two-PR stack. The following PR connects these primitives to the flash and bubble services and trigger controls.

Validation: the complete stack passed 7,644 tests (1 skipped), built successfully, and the owner confirmed playback and niche rotation during a desktop test. Focused tests cover delivery geometry and pool consumption/invalidation.

Playback is bounded to eight live decoders, 480 px and 15 fps; posters cover startup and budget fallback.

Standalone foundation build passed with zero errors. Integration PR: #1426.


CodeBambi:
Reviewed for the 6.10.1 pre-release. Rendering primitives for clip-backed flash and bubble faces. Read the Dispose path in Services/Media/FlashClipPlayer.cs specifically, since that is where the native trap lives: Stop() then Dispose() then FreeHGlobal, all on a background task, so the vmem callback buffer is retained until playback has actually stopped, and the budget slot is kept rather than double-released if cleanup throws. Caps are the decided ones: 8 simultaneous players, 480 px longest edge, ~15 fps via the 66 ms gate in Display. Muted, hardware decoding off, frames frozen before they cross to the dispatcher. CI green. Merging.

PR #1428, opened 2026-09-19
fix(launcher): every game tile fits, and the way back is labelled

CodeBambi:
Tester feedback on the v6.10.0 launcher, two items.

## Cards do not fit

The games column was a `ScrollViewer` over a 2-column `UniformGrid` whose tiles had a fixed 156 px art row, so six entries showed two rows at the window's MinHeight and hid the third below the fold.

- `Services/Launcher/LauncherGridLayout.cs` (new, pure): `Columns(width, count)` (2, or 3 from 740 px), `Rows`, `TileHeight`, `ArtHeight` (96 px floor), `NeedsScroll` / `GridHeight` (252 px card floor).
- `LauncherWindow.xaml.cs` `FitTiles()`: on the column's `SizeChanged` and after `BuildTiles`, sets the grid's columns from the column's ActualWidth and its height to the scroller's viewport, so the rows share the height. Only when a row would drop under the floor does the grid grow past the viewport and the scroller scroll.
- `LauncherWindow.Tiles.cs`: the art row is `*` with `MinHeight 96` instead of a fixed 156; text margins and the Play button tightened by a few px (part of the floor's sum). Tilt, glow, rim, the mystery "?" plate and the race tile's 330x200 Viewbox road all scale with the row and are untouched.
- Numbers: the default 1280 x 800 window leaves the column 756 px, so the threshold sits at 740 and six tiles land as 3 x 2 at ~329 px a row, nothing to scroll. The 1000 x 680 minimum window is 2 x 3 at ~179 px a row, under the floor, so it keeps the tiles at 252 px and scrolls.

## The way back from CCP

`BtnBackToLauncher` existed as a bare chevron and nobody found it.

- `MainWindow.xaml`: the button now reads chevron + "CC Labs" (new key `launcher_back_to_client_short`, all 9 locales, edited as lines, no BOM). Width is auto with a 12 px pad, so the window controls to its right keep their spots.
- `TrayIconService`: a "Back to CC Labs" row (`launcher_back_to_client`) under "Show the window", calling `LauncherHost.BackToLauncher()`; visible only while `LauncherHost.SurfaceInPlay` (launcher created, or the boot did not ask for the panel outright), greyed under Lockdown.
- `LauncherHost.SurfaceInPlay` is new; the title-bar button itself has no visibility rule and never had one, so the tray row mirrors "the launcher is part of this run" rather than a rule that did not exist.

## Tests

- `Tests/ConditioningControlPanel.Tests/LauncherGridLayoutTests.cs`: columns switch at 740 and never exceed the count, rows round up, the default window fits without scrolling, the minimum window falls back to the floor, zero rows give zero.
- Build 0 errors, 0 warnings. Suite 7652 passed / 0 failed / 1 skipped (the pre-existing JackpotRemix spike skip).

## Not verified in the running app

Layout is verified by arithmetic and a clean build only: the star row under a UniformGrid cell, the 3-column blurb wrap at the default width, the tray row and the labelled button are owed a desk look.


CodeBambi:
Reviewed: diff read, no em-dashes, build 0 errors, suite 7652/0. Dry merge with #1429 is conflict-free. Landing on the owner's word.

PR #1429, opened 2026-09-19
feat(launcher): labelled Media button, asset presets, mod switcher

CodeBambi:
## What changed

Tester feedback on the v6.10.0 launcher (`Windows/Launcher/*`).

**1. Labelled Media button.** The media picker hid behind a bare gear. Same title-bar spot, now a pill with a small icon and the word "Media" (`LauncherBarPill` style, `BtnGear_Click` unchanged, still opens "Media and sound").

**2. Asset preset in the Media dialog.** New "Library" section: an Asset preset ComboBox (active preset selected, applies on change, `Foreground="Black"` on box and rows) and an "Open asset browser" link that closes the dialog and opens the panel on the assets tab through `LauncherHost.OpenPanelTab("assets")`.
- The write is factored into `Services/AssetPresetService` (pure: `Active`, `Find`, `Apply`), so it works with the panel created but tray-hidden.
- The panel's own preset combo (`MainWindow.Assets.cs`) now goes through the same `Apply` and then does its repaint; the launcher calls `MainWindow.ApplyAssetPresetFromLauncher`, which is the same repaint plus moving the panel's combo. One write, two callers.

**3. Mod switcher in the title bar.** A "Mod: <active mod>" pill; a click drops a menu of the installed mods (panel's order: stock canonical, then user mods A-Z, via pure `LauncherModMenu`), active one ticked, "Manage mods" as the last row.
- A pick calls `MainWindow.SwitchActiveModFromLauncher`, which is the top-bar combo's own two steps (`ActivateMod` + `ApplyActiveModChange`). No second activation path.
- "Manage mods" goes through `LauncherHost.OpenPanelModManager`: the panel comes up first (the Mod Manager is a modal owned by it), then `BtnManageMods_Click` runs, same launcher the combo's footer row uses. `LauncherHost.OpenPanel` gained a private continuation for that.
- The pill follows `ModService.ModChanged`, so the panel's combo, the Mod Manager and the first-run picker all update it.

8 new loc keys (`launcher_media_button`, `launcher_mod_label`, `launcher_mod_tooltip`, `launcher_mod_manage`, `launcher_media_library`, `launcher_media_preset`, `launcher_media_preset_hint`, `launcher_media_open_browser`) in all 9 language files, edited line by line, no BOM.

## Why

Testers could not find the media picker, could not switch asset presets without opening the panel, and had no way to see or change the active mod from the launcher.

## Tests

- New `LauncherModMenuTests`: mod ordering (stock first, user A-Z, unknown built-ins kept, null safe), pill label formatting, `AssetPresetService.Apply` (replaces the disabled set with a copy, remembers the id, unknown id writes nothing) and `Active` (current id, else default row).
- Build 0 errors / 0 warnings. Suite: 7640 passed, 0 failed, 1 skipped.

## Caveats

Not run in the app. Owed a desk look: pill widths beside the account chip at 1000px min width, the dark theme's implicit ContextMenu under the Mod pill, the preset combo's closed-box text, and Manage mods opening over the panel after the launcher fade.


CodeBambi:
Reviewed: diff read, no em-dashes, build 0 errors, suite 7640/0, one apply path for presets and mods. Landing on the owner's word.

PR #1430, opened 2026-09-20
fix(backroom): the card table lets the SP chip go, and reduced motion keeps the settle

CodeBambi:
Two bugs found chasing the tester report "my SP total jumps between 0 and the real total" at the blackjack table.

**The chip pinned at zero.** `stations/cards/station.js` `chip.show()` coerced before it tested, and `Number(null)` is 0 and passes `Number.isFinite`. THE BANK's release call is `show(null)`, so instead of handing the chip back to the room it pinned it at 0. `flightValue()` in `stations/cards/hand.js` now tests `typeof v === 'number'` first.

Worth watching for elsewhere: several `Number.isFinite(Number(x))` guards in the room read an explicit JSON `null` as 0.

**Reduced motion killed the settle.** The card and roulette banks played `run.step(startMs)` - a frame - where `play()` wants an event array, so the settle threw "events is not iterable". Only `skip()` returns a bare array. A source guard in `shared/win/tests/bank.test.mjs` now holds all four station banks.

Back Room unit gate 832/833 (the one red is the orphan `br_station_breakout` lexicon row left by the Breakout exclusion).

Note: #1399 carries a defensive guard for the same symptom; this is the root cause. Both can stand.


CodeBambi:
CI is red on ConditioningControlPanel.Tests.UiThreadStackCaptureTests.ASelfSnapshotYieldsTheCallingThreadsManagedFrames, which is a managed stack-frame capture test and cannot be reached by this diff: the change is four .js/.mjs files under Resources/web/backroom and nothing else. It passes locally - full C# suite on the merged tip a4df4b733 is 7674 passed / 0 failed / 1 skipped, that test included, and the Back Room node gate is 842/843 with the one red being the known orphan br_station_breakout row from the Breakout exclusion. Treating the CI red as a flake in the runner and merging for the 6.10.1 pre-release.

PR #1426, opened 2026-09-19
Deliver animated bubble faces and fix Scrolller flash rotation

CodeBambi:
Flash bubbles now show animated media inset inside the glass with a soft edge. Popping one delivers that picture at the bubble position and fires the configured CCP burst count, size, and duration. Pull is retired and the Melt toggle appears beside the other trigger types.

Scrolller flashes now request actual silent clips instead of GIF-feed posters. Ready clips are consumed, refills sample several channels, and niche changes clear old selections and reject stale downloads. Real online GIF/WebP frames are also preserved.

Stacked on #1425. Merge #1425 first, then retarget this PR to main. The final tree is identical to the owner-tested commit d130947a3; only the review split changed.

Validation: 7,644 tests passed, 1 skipped; final build passed. Owner confirmed playback and niche rotation in the relaunched desktop app. Eight live clip decoders maximum; additional content uses posters.

PR #1398, opened 2026-09-18
fix(backroom): desk look (pointer lock + arrows), Play prompt, Stations list instead of the map

CodeBambi:
Desktop tester feedback on the Back Room 3D casino, three items in one PR. Worktree `fix/br-desk-look` off `origin/main`.

## 1. Look around: mouse look (pointer lock) + arrow keys

**Feedback:** "click and drag to look around is terrible... tie mouse movement to looking around like a proper first person game or give me keyboard keys".

**Now:**
- A mouse click on the canvas while walking requests pointer lock; mouse movement then drives yaw/pitch (`LOOK_SENSITIVITY = 0.0022` rad/px, `scene.js`). Esc frees the mouse; the next click takes it again. Nothing ever re-captures on its own.
- The capturing click is not a tap, so a click to get the mouse back cannot walk you into whatever the crosshair rested on. Under lock, a click picks at the crosshair (the cursor is hidden and its position frozen).
- Click-drag stays as the fallback when the lock is unavailable or refused (after `LOCK_REFUSALS = 2` refusals in a row the room stays on drag + tap). Fingers and pens never lock, so `touch-control.js` is untouched.
- Arrow keys look at a steady rate (`KEY_YAW_RATE = 1.8`, `KEY_PITCH_RATE = 1.2` rad/s); WASD walks; Q/E untouched (E is the visit). Arrows no longer strafe/walk.
- The lock is released whenever a station takes the screen (`prepareVisit`, `seat`, `hold`), a card opens (memorabilia viewer, welcome placards), Room Service opens, the M list opens, or the room pauses/halts.
- The Esc that frees the mouse is not the Esc that leaves the room: `scene.lookJustFreed()` (locked, or unlocked within 500 ms) short-circuits the Escape handler in `main.js`.
- Hint text is now `br_look_hint`: "Click to look, WASD to walk, arrows to turn, E to play, Esc to free the mouse".

## 2. A prompt on interactable objects

**Feedback:** "the only way to tell is to walk up and press E to test".

**Why:** `hud.nearest(row)` only wrote the station's name into the walk hint's `title` (a tooltip on a 12 px line at the foot of the screen).

[...]

PR #1397, opened 2026-09-18
feat(backroom): the Annex, a second 3D room with the race door and two locked doors

CodeBambi:
## Tester feedback

> if you're going to run with the 3D thing then turn this into a hub for all the 3d bits, starting with the racing game. straight up add a second room to this that has a door to the racetrack, and maybe leave another door or two in there locked to open in future content

## The Annex

A second room behind the **west wall** of the Back Room. That wall carries the fewest fixtures: the wheel plinth ends at z -0.06 and the rose slot's body starts at z 2.05, so a 1.8 m doorway at z 0.1..1.9 touches nothing (the west-wall media screen hangs above y 2.14, so the doorway is 2.05 tall and the screen stays uncovered). The counter at -z and the entrance at +z are untouched.

- `room/annex.js` (new): procedural, no new GLBs or textures. Floor in the casino's own spiral shader material, plum walls, brass jambs, neon skirting in the venue palette, two of the scene's point lights, and a canvas "The Annex" blade sign on a brass bracket by the doorway. The wall cut is a clip box on the shell's wall materials (`renderer.localClippingEnabled`, `clipIntersection`), so `shell.glb` is never edited; a lacquer lining and brass trim dress the opening.
- Three doors, first-class E targets with the same prompt/nearest logic as stations (`ANNEX_PORTALS` in `walk.js`, appended to `stationRows`; taps on a door work through `fixtureOf`):
  - **Racing Thoughts** (far west wall) is LIVE. E (or a tap) hands the row to `main.js visit()`, which routes `portal === 'race'` to the existing `openRace()` - the same `race-portal.js` flow as the cabinet, so the return pose and `RACE_OWNERSHIP_KEY` keep working. `validatedRoomPose` now accepts poses inside the annex/doorway, so coming back from the race lands you in front of the race door.
  - Two LOCKED doors (north and south walls): padlock shape and a "Coming soon" plaque. E rattles the padlock and blinks the plaque (`annex.refuse`); no HUD line needed, so `hud.js` is untouched.
- `walk.js` (pure): `ANNEX`, `DOORWAY`, `inAnnex`, `inHall`, `ANNEX_PORTALS`; `blocked()` is the union of the main rectangle and the annex joined by the doorway gap; `reachable()` grids the union while keeping its main-room grid points.
- `scene.js` / `main.js`: additive only (import, build, rows, `fixtureOf`, `update`, `dispose`, `annex` on the scene API; two lines in `visit()`), to stay out of the mouse-look lane's way.
[...]

PR #1431, opened 2026-09-20
release: 6.10.1 "Let the games begin"

CodeBambi:
Version bump for the 6.10.1 pre-release, on top of everything merged today: the five Back Room desk-feedback lanes (#1395 #1396 #1397 #1398 #1399), the clip stack (#1425 #1426) and the card table SP chip fix (#1430).

- csproj, UpdateService (AppVersion + patch notes), installer.iss, build-installer.bat
- update button keys in MainWindow.xaml plus the two new loc keys in nine languages
- drops the orphan `br_station_breakout` row from all nine locales: #1422 took the station out of stations.json and the publish tree but left its lexicon row, which the Back Room lexicon gate reads as a row nothing asks for

Content packs are **not** rebuilt. `DeriveCycleTag` maps 6.10.1 to the `v6.10.0` cycle, so this build fetches the seven packs and the manifest already attached to that release. The tag only needs the installer.

Verified in a throwaway worktree at the merged tip: build 0 errors, C# suite 7674 passed / 0 failed / 1 skipped, Back Room node suite 845/845 (was 842/843 before the orphan row came out).


CodeBambi:
Reviewed on the pushed tip `eef78968c`. Two commits landed on this branch since it was opened:

- `c4b7beb4d` fix(launcher): landscape card art. The Arcademy tile had never had a picture - its art path pointed into `Resources/web/`, which ships as Content on disk while `ModResourceResolver` resolves through a `pack://` uri that only reaches a compiled WPF Resource, so it drew the glyph plate every launch with nothing in the log. The art moved to `Resources/features/`, the plate is a 16:9 landscape sized off the card's width so no picture is cropped, and the text block is a fixed height so a one-line blurb and a two-line blurb no longer put their titles on different lines. `LauncherGridLayout` owns the arithmetic and its suite was rewritten with it.
- `eef78968c` release: one addendum line in the patch notes for the above, written through to `UpdateService.CurrentPatchNotes`.

Verified on the tip: build 0 errors, suite **7677 passed / 0 failed / 1 skipped**, launcher tests 119/119. Version strings all read 6.10.1 in the five places. Installer built, signed (`CN=CC Labs Srls`, timestamped), installed and run at the desk: the launcher opens on six landscape cards with the Arcademy art in place.

Content packs are unchanged: `DeriveCycleTag("6.10.1")` is `v6.10.0`, so the packs stay attached to that release and this tag carries the installer only.

PR #1436, opened 2026-09-20
Breakout 04/16: Reduce word-effect rendering cost

CodeBambi:
Reduces repeated fullscreen work in DROP, SINK, RELAX and LET GO while preserving their gameplay effects and reduced-motion behavior.

Stack layer 4/16. Base: `codex/breakout-03-performance`. Merge in order and retarget after each base lands.

Validation: existing Breakout, word-effect, Loom and music-option tests pass at this layer; complete stack passes241 tests. Changed lines: 461. Owner browser listening/visual check remains pending.

Preserves the original uncommitted working tree. Office ending is a separate follow-up.


CodeBambi:
Contained in #1448, merged to main as edc73deb5.

PR #1457, opened 2026-09-20
fix(breakout): the pause pill reads Pause

CodeBambi:
The pause button's glyph had been through a utf-8 round trip twice and drew
30 characters of Latin-1 soup across the top-left of the HUD. Caught in a
browser while wiring the web preview.

It is `&#9208;` now, an HTML entity in the same template that already writes
the Start button's arrow as `&#9656;`, so no encoding pass can reach it again.

One line. Breakout node suite 202 passed, 0 failed.

https://claude.ai/code/session_01PkjA6aWZwsEKH6Bj6PJdnA


CodeBambi:
Reviewed. One line, in an HTML template that already writes a glyph the same way three lines above. Windows CI green; Breakout node suite 202 passed. Found in a browser, not by reading.

